PR #33, opened 2023-10-13
Cleanup README, Move some instructions to docs/DEV.md

liruilong940607:
Also add a workflow for auto pypi. It will be triggered once we create a new release on main branch from here:

Screenshot 2023-10-12 at 7 00 12 PM

It assumes the repo has an environment called `production` and a secret token `PYPI_TOKEN` being setup in the repo setting. @vye16 

And worth to note only @vye16 will be able to trigger this because it recognize her username and password, and the package will be uploaded to her PYPI project.


liruilong940607:
This is what the new README looks like:

Screenshot 2023-10-12 at 7 07 23 PM

vye16 (approved):
Looks great!

PR #41, opened 2023-10-13
num_sh_bases() is used in both host and device

liruilong940607:
Fix the following bug when compiling:

```
gsplat/cuda/csrc/bindings.cu(72): error: calling a __device__ function("num_sh_bases(unsigned int)") from a __host__ function("compute_sh_forward_tensor") is not allowed

gsplat/cuda/csrc/bindings.cu(104): error: calling a __device__ function("_Z12num_sh_bases1?") from a __host__ function("compute_sh_backward_tensor") is not allowed
```


maturk (approved):
You are right. Thanks for the fix.

PR #47, opened 2023-10-13
Workflow for all permutation building test

liruilong940607:
The Windows compiling is finally passed: https://github.com/nerfstudio-project/gsplat/actions/runs/6512955506/job/17691629348

Here to restore workflow for all permutation building test.


maturk (approved):
Ty for the help!

PR #53, opened 2023-10-17
remove point-wise init in simple_trainer.py.

imlixinyang:
Remove slow point-wise initialization in simple_trainer.py. 
The result has been tested to be the same with the original one.


maturk (approved):
LGTM! Thanks for the improvement.

PR #63, opened 2023-10-25
Fix test_project and test_rasterize

maturk:


Zhuoyang-Pan (commented):
Great! I think we need to take a deeper look into the CUDA kernel to figure out the index issue now. I also notice that there are many idx ifs in the PR. Are they for debugging purposes?  It would be great if we could clean them before finalizing the PR.

I've also encountered some inconsistencies with the tests for `test_get_tile_bin_edges` and `test_map_gaussians`. They pass on the GitHub tests but fail on my local setup. It seems that we also need to add cx, cy in `_torch_impl.project_gaussians_forward` in those two tests?

maturk:
closing this since it is way out of date and haven't had the time to fix anything :)

PR #67, opened 2023-11-01
CUDA optimizations

kerrj:
A variety of speedups, including
- Warp-aggregated atomic adds in rasterize backward to reduce the number of global atomics (using cg::reduce and writing to global memory only from the rank 0 thread in each warp)
- Refactor some shared memory ops to use float3 instead of float2 + float
- refactoring SH kernels to accept a `degrees_to_use` parameter to avoid a contiguous tensor memory copy
- Changing all relevant pointers to `__restrict__`

I played with packing some global memory vectors into float4, (rgb+opacity->rgba, conics from float3 to float4), but didn't see very much speed improvement at all. This is a bit surprising given float4 reads/writes compile to wide instructions, and are better cache aligned (nsight compute does show much fewer uncoalesced L2 global reads, but this doesn't significantly increase speed....) If someone has experience profiling/optimizing CUDA, input is very much appreciated.


setup.py
```diff
@@ -72,7 +72,7 @@ def get_extensions():
 
     nvcc_flags = os.getenv("NVCC_FLAGS", "")
     nvcc_flags = [] if nvcc_flags == "" else nvcc_flags.split(" ")
-    nvcc_flags += ["-O3"]
+    nvcc_flags += ["-O3","--use_fast_math","-lineinfo"]
```
vye16, line 75: can we make this an option/flag in cli?

gsplat/cuda/csrc/forward.cu
```diff
@@ -277,7 +277,11 @@ void bin_and_sort_gaussians(
 
     // sort intersections by ascending tile ID and depth with RadixSort
     int32_t max_tile_id = (int32_t)(tile_bounds.x * tile_bounds.y);
+#if defined(WIN32) || defined(_WIN32) || defined(__WIN32) && !defined(__CYGWIN__)
```
vye16, line 280: thanks for dealing with the different os builtins, I didn't think of this

vye16 (approved):
looks great! only thing is to make the compile with lineinfo opt in and not opt out

gsplat/cuda/csrc/forward.cu
kerrj: I think this is from some other PR actually... I'm not sure why its showing up here

PR #78, opened 2023-12-05
Unnecessary CUDA code cleanup

liruilong940607:
A big cleanup mainly to remove the unnecessary cuda code, includes:

- `_C.bin_and_sort_gaussians`: The CUDA implement of this function is removed. It can be implemented on the python side by combining `_C.map_gaussian_to_intersects` + `torch.sort` + `_C.get_tile_bin_edges`. The performance is overall similar. Note `torch.sort` automatically choose radix sort for large input and partial heap sort for small input, so now it actually runs slightly faster than before with small number of gaussians.

- `_C.compute_cumulative_intersects`: This CUDA implementation is removed and replaced by `torch.cumsum` which also implemented warp sum under the hood so has similar performance. The benefit is that we now do not rely on CUB so can be compiled with CUDA < 11.

- `rasterize.cu`: (ND) Rasterize is now implemented in python using the util functions from `forward.cu` and `backward.cu` so this file is not needed thus removed.

- Cleanup CUDA function namings: `***_kernel`, `***_tensor`, `***_kernel_tensor`, `***_impl` etc.

- A proposal: calling `BinAndSortGaussians.apply(***)` is a bit uncommon way running a function comparing to simply call `map_gaussian_to_intersects(***)`. So I propose to turn those classes into functions, which I have added a few in this PR. We can maintain backward compatibility and raise a warning when user uses the class version.


liruilong940607:
`tests/*` and `examples/simple_trainer.py` are tested to be unaffected by this PR

gsplat/cuda/csrc/backward.cuh
```diff
@@ -5,70 +5,68 @@
...
-    const float *background,
-    const float *final_Ts,
-    const int *final_index,
-    const float *v_output,
-    float2 *v_xy,
-    float3 *v_conic,
-    float *v_rgb,
-    float *v_opacity,
-    float *workspace
+    const int32_t* __restrict__ gaussians_ids_sorted,
```
brentyi, line 36: do we want to flip all of the `int` types to `int32_t`? (like `final_index`)

brentyi (commented):
looks reasonable to me!

PR #85, opened 2023-12-16
Project gaussians: adding numerical tests, resolving gradient bug

vye16:
* added numerical tests for forward and backward passes against autograd on torch_impl in `tests/test_project_gaussians.py`
* fixed a gradient bug in projection. NOTE this changes the interface for the `ProjectGaussians` function, which now expects `viewmat` and `projmat`, rather than `viewmat` and the full matrix of`projmat * viewmat` as it did before.
* forward pass matches up to 1e-5 precision, backward pass matches up to 5e-4 precision.


liruilong940607:
All looks great. I think the change on the interface of `ProjectGaussians` makes sense but we might need to consider backward compatibility. Is there a way to find out if the user is passing in the `fullmat` as before and raise a warning (preferable option)? Another option is to remain the API unchanged but compute the `projmat` from `fullmat` and `viewmat`.

liruilong940607 (approved):

PR #97, opened 2024-01-05
Remove redundant projection matrix

oseiskar:
The projection matrix is a redundant concept here. The code assumed it was computed in a particular way from fx, fy, cx, cy and other parameters which eventually canceled out.

This PR simplifies the code by removing the projection matrix, but leaves the `projmat` parameter to `ProjectGaussians` for backwards compatibility. If merged & released, this PR can simplify Nerfstudio [like this](https://github.com/SpectacularAI/nerfstudio/commit/bd234898725b1df46984a8f8b0925e2de43d4581)

EDIT: `_torch_impl` was now also changed to match the new implementation to fix the unit tests (i.e., check correctness of the gradients)


gsplat/cuda/csrc/helpers.cuh
```diff
@@ -104,36 +110,26 @@ inline __device__ float4 transform_4x4(const float *mat, const float3 p) {
...
-    float3 p_proj = {p_hom.x * rw, p_hom.y * rw, p_hom.z * rw};
-    return {
-        ndc2pix(p_proj.x, img_size.x, pp.x), ndc2pix(p_proj.y, img_size.y, pp.y)
-    };
+    float rw = 1.f / (p_view.z + 1e-6f);
+    float2 p_proj = { p_view.x * rw, p_view.y * rw };
+    // TODO: check if +-0.5 should be added to pix coords or not
+    // this depends on rasterize
+    float2 p_pix = { p_proj.x * fxfy.x + pp.x, p_proj.y * fxfy.y + pp.y };
+    return p_pix;
```
oseiskar, line 120: This may slightly change the results: the current code subtracts `-0.5f` from the pixel coordinates, which is not done here. It can be added back but is there a good reason to do so? If fx, fy ppx, ppy originate from e.g., COLMAP, the pinhole projection formula does not include such a term.

gsplat/cuda/csrc/helpers.cuh
```diff
@@ -104,36 +110,26 @@ inline __device__ float4 transform_4x4(const float *mat, const float3 p) {
 }
 
 inline __device__ float2 project_pix(
-    const float *mat, const float3 p, const dim3 img_size, const float2 pp
+    const float2 fxfy, const float3 p_view, const float2 pp
```
oseiskar, line 113: The API of this internal helper is also changed: In the previous version, the parameter was assumed to be a (model-)view-projection matrix instead of a projection matrix, and `p` was in world coordinates. Now the transformation to camera coordinates is handled separately and this method only performs the projection, for better separation of concerns.

oseiskar:
Rebased on latest main and fixed conflicts. Compatible with https://github.com/nerfstudio-project/nerfstudio/pull/2888

oseiskar:
Since this still seems to be confusing, I'll try to explain in more detail: The current code in Nerfstudio is equivalent to this
```
project_gaussians(
    intrinsics,
    projmat = f(intrinsics, other_parameters),
    other_parameters,
    more_parameters)
```
and if one would independently modify either projmat or intrinsics without changing the other, the method would break. It's therefore quite clear to me that it would be better to simplify to either of the two options:
 1. `project_gaussians(intrinsics, other_parameters, more_parameters)` (implemented here)
 2. OR `project_gaussians(projmat, other_parameters, more_parameters)`

where `project_gaussian` would internally compute either `projmat` or `intrinsics`, if it needs them (it actually does not, in the case of `projmat`).

The question of which one to retain, projmat or intrinsics, is also rather clear: the correct answer is intrinsics (1), because it allows extending the code to use other camera models than the ideal undistorted pinhole model.

Why `projmat` is there in the first place is probably legacy from experimenting with things that are closer to standard OpenGL rendering where NDC coordinates are a relevant concept and not wanting to touch it once it's working. However, it's now a legacy roadblock that complicates and prevents certain types of extensions to the method.

jb-ye:
It looks good to me. One thing that needs to be double checked is to run test_project_gaussians.py locally and verify the gradient calculation is correct, especially after fixes #136 , #139 and #140 .

oseiskar:
> It looks good to me. One thing that needs to be double checked is to run test_project_gaussians.py locally after rebasing and verify the gradient calculation is correct, especially after fixes #136 , #139 and #140 .
> 
> @oseiskar Could you rebase and test this PR?

Done! The gradient tests now passed on my machine.

My original implementation of the tests actually had a bug in using the `vjp` utility. This was revealed by the random `viewmat`, which wasn't used in my old version. Fixed it [like this on the original branch](https://github.com/SpectacularAI/gsplat/commit/2b7c251eea2b2e809f13a42812b4cf58c13a8885).

Also opened a corresponding pull request in Nerfstudio https://github.com/nerfstudio-project/nerfstudio/pull/3019

gsplat/cuda/csrc/helpers.cuh
vye16: yes, apologies. for the sake of laziness, we compute the full matrix, proj_mat @ view_mat and pass that matrix around. we pass both this full matrix and view matrix around because both are necessary for the backward pass. this new API works well too.

gsplat/project_gaussians.py
```diff
@@ -15,7 +15,7 @@ def project_gaussians(
     glob_scale: float,
     quats: Float[Tensor, "*batch 4"],
     viewmat: Float[Tensor, "4 4"],
-    projmat: Float[Tensor, "4 4"],
+    projmat: Optional[Float[Tensor, "4 4"]],
```
vye16, line 18: I get that this is good for backward compatibility, but if we're going to phase out the projection matrix, we should ultimately drop this argument.

vye16 (approved):
thank you for testing with numerical tests -- if they now pass, I'm good with merging this in. I also am hesitant about keeping around the deprecated projmat argument (`examples/simple_trainer.py` should also be updated) but for now, I'm okay with this.

gsplat/cuda/csrc/helpers.cuh
jb-ye: @oseiskar I think we really need to add this back. Otherwise it would bring quality degradation in splatfacto. For an image, each pixel's index (i,j) actually represents the area of intensity whose center 2d point is at location (i+0.5, j+0.5).

oseiskar: Commented there https://github.com/nerfstudio-project/gsplat/pull/151#issuecomment-2027978449

PR #102, opened 2024-01-16
Test and example trainer updated

hardikdava:
- using latest apis in example trainer and tests


liruilong940607 (approved):
Thanks for the update. LGTM

PR #106, opened 2024-01-22
Fix build detection on Windows

pierotofy:
Hello :wave:,

the current code on Windows will fail to detect when a build has already been done, due to libraries having .lib extension rather than .so and will repeat the build at every program load.

The proposed change checks for .lib extension also.


liruilong940607 (approved):
LGTM!

PR #117, opened 2024-02-01
Compute density compensation for screen space blurring of tiny gaussians

jb-ye:
# Why
Current gaussian splatting (both Inria and nerfstudio) doesn't sufficiently deal with the case when rendering tiny gaussians at substantial different resolution than the captured one. The reason is caused by the [0.3 pix kernel screen space blurring](https://github.com/nerfstudio-project/gsplat/blob/main/gsplat/_torch_impl.py#L193-L194) applied to the tiny splats. 

A tiny splat that has been enlarged by 0.3 pixel can block more gaussians behind it if rendering at lower resolution than the captured one, or looks much thinner than it should be if rendering at higher resolution than the captured one. One would easily observe this artifacts by changing the distance or resolution of rendering. 

This issue is [acknowledged](https://github.com/graphdeco-inria/gaussian-splatting/issues/294#issuecomment-1772688093) by the original author of Gaussian splatting, and was [addressed](https://github.com/wanmeihuali/taichi_3d_gaussian_splatting/issues/151) in another open source repo. 

Addressing this issue has significant benefit in an interactive webviewer, where one zoom in/out to inspect a splat model, and see much less artifacts then before. Another simple quantitative experiment one can do is to train at 1/2 resolution and evaluation at the original resolution, addressing the issue expects to improve metrics. 

# Solution

The solution is rather simple. We will compute a compensation factor $\rho=\sqrt{\dfrac{Det(\Sigma)}{Det(\Sigma+ 0.3 I)}}$ for each splat and multiply it to the opacity of gaussian before rasterization. The same solution was also proposed in [another research paper](https://github.com/autonomousvision/mip-splatting).

The PR will basically modify the output of rasterize_gaussians to return the compensation factor $\rho$ and  the change allows us to have two modes in nerfstudio's splatfacto: the classic mode which mimic the behavior of official GS, and the new mode which address the "aliasing-like" issue. 

See results [here](https://github.com/jb-ye/gsplat/issues/1) (I had to upload the video to my own forked repo's issue due to file size limit).

### DIFF
--- a/gsplat/cuda/csrc/forward.cu
+++ b/gsplat/cuda/csrc/forward.cu
@@ -26,6 +26,7 @@ __global__ void project_gaussians_forward_kernel(
     float* __restrict__ depths,
     int* __restrict__ radii,
     float3* __restrict__ conics,
+    float* __restrict__ compensation,
     int32_t* __restrict__ num_tiles_hit
 ) {
     unsigned idx = cg::this_grid().thread_rank(); // idx of thread within grid
@@ -61,8 +62,11 @@ __global__ void project_gaussians_forward_kernel(
     float cy = intrins.w;
     float tan_fovx = 0.5 * img_size.x / fx;
     float tan_fovy = 0.5 * img_size.y / fy;
-    float3 cov2d = project_cov3d_ewa(
-        p_world, cur_cov3d, viewmat, fx, fy, tan_fovx, tan_fovy
+    float3 cov2d;
+    float comp;
+    project_cov3d_ewa(
+        p_world, cur_cov3d, viewmat, fx, fy, tan_fovx, tan_fovy,
+        cov2d, comp
     );
     // printf("cov2d %d, %.2f %.2f %.2f\n", idx, cov2d.x, cov2d.y, cov2d.z);
 
@@ -88,6 +92,7 @@ __global__ void project_gaussians_forward_kernel(
     depths[idx] = p_view.z;
     radii[idx] = (int)radius;
     xys[idx] = center;
+    compensation[idx] = comp;
     // printf(
     //     "point %d x %.2f y %.2f z %.2f, radius %d, # tiles %d, tile_min %d
     //     %d, tile_max %d %d\n", idx, center.x, center.y, depths[idx],
@@ -372,14 +377,16 @@ __global__ void rasterize_forward(
 }
 
 // device helper to approximate projected 2d cov from 3d mean and cov
-__device__ float3 project_cov3d_ewa(
+__device__ void project_cov3d_ewa(
     const float3& __restrict__ mean3d,
     const float* __restrict__ cov3d,
     const float* __restrict__ viewmat,
     const float fx,
     const float fy,
     const float tan_fovx,
-    const float tan_fovy
+    const float tan_fovy,
+    float3 &cov2d,
+    float &compensation
 ) {
     // clip the
     // we expect row major matrices as input, glm uses column major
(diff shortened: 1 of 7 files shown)


gsplat/project_gaussians.py
```diff
@@ -47,12 +47,13 @@ def project_gaussians(
...
 
     Returns:
-        A tuple of {Tensor, Tensor, Tensor, Tensor, Tensor, Tensor}:
+        A tuple of {Tensor, Tensor, Tensor, Tensor, Tensor, Tensor, Tensor}:
 
         - **xys** (Tensor): x,y locations of 2D gaussian projections.
         - **depths** (Tensor): z depth of gaussians.
         - **radii** (Tensor): radii of 2D gaussian projections.
         - **conics** (Tensor): conic parameters for 2D gaussian.
+        - **compensation** (Tensor): the density compensation for blurring 2D kernel
```
liruilong940607, line 56: This extra return would break backward compatibility. Personally I'm fine with it as we are in active-developing version `0.1.x`. But I'll let @vye16  to decide starting from when we want to maintain backward compatibility.

gsplat/cuda/csrc/forward.cu
```diff
@@ -88,6 +92,7 @@ __global__ void project_gaussians_forward_kernel(
     depths[idx] = p_view.z;
     radii[idx] = (int)radius;
     xys[idx] = center;
+    compensation[idx] = comp;
```
liruilong940607, line 95: Read/Write the global memory is usually the most time consuming part in a kernel (computation is usually not the burden). I tested this a bit and it slows down the `project_gaussians` from 3000 it/s to 2800 it/s which is not that much so I think is fine. Especially that `project_gaussians` is much cheaper comparing to the `rasterization` stage. I'm fine with this tiny little extra burden but just want to point it out for future reference.

liruilong940607 (commented):
I'm fine with merging this PR but I want to let @vye16 to decide if we need backward compatibility or not.

gsplat/cuda/csrc/forward.cu
liruilong940607: The code I used to test this.

```python
import torch

def profiling(N: int = 1000000, D: int = 3):
    import tqdm

    from gsplat import project_gaussians, rasterize_gaussians

    torch.manual_seed(42)
    device = torch.device("cuda:0")

    means3d = torch.rand((N, 3), device=device, requires_grad=False)
    scales = torch.rand((N, 3), device=device) * 5
    quats = torch.randn((N, 4), device=device)
    quats /= torch.linalg.norm(quats, dim=-1, keepdim=True)

    viewmat = projmat = torch.eye(4, device=device)
    fx = fy = 3.0
    H, W = 256, 256
    BLOCK_X = BLOCK_Y = 16
    tile_bounds = (W + BLOCK_X - 1) // BLOCK_X, (H + BLOCK_Y - 1) // BLOCK_Y, 1

    pbar = tqdm.trange(10000)
    for _ in pbar:
        xys, depths, radii, conics, num_tiles_hit, cov3d = project_gaussians(
            means3d,
            scales,
            1,
            quats,
            viewmat,
            projmat,
            fx,
            fy,
            W / 2,
            H / 2,
            H,
            W,
            tile_bounds,
        )
```

liruilong940607:
Needs to run  the formatter `black . gsplat/ tests/ examples/` for passing the core tests.

hardikdava:
@jb-ye Really great improvement. Amazing work @jb-ye Can't wait to try it out.

jb-ye:
> Needs to run the formatter `black . gsplat/ tests/ examples/` for passing the core tests.

I used a newer version of black. Now updated black with the same version used by devops and re-run the formatter.

gsplat/cuda/csrc/forward.cu
jb-ye: Thanks for the evaluation effort.

gsplat/project_gaussians.py
jb-ye: Hi @vye16 , could you help take a look at this PR and see if you have any other comments other than @liruilong940607

f-dy:
Initial discovery of the issue:
- https://github.com/graphdeco-inria/gaussian-splatting/issues/294#issuecomment-1772688093

Example showing the effect of the fix:
- https://github.com/graphdeco-inria/gaussian-splatting/issues/294#issuecomment-1777782320

vye16 (approved):
Looks good to me, aside from this backward compatibility issue. How much faster is it do this in the kernel than with the 2d covariance returned in the original projection function? If it's much faster, I'm happy to break backward compatibility.

PR #118, opened 2024-02-01
updating project gaussians torch impl and numerical tests

vye16:
* update the project gaussians pytorch impl to match the existing cuda backend for numerical testing purposes.
* add gradient numerical tests for project gaussians. Note the current backend gradients don't match with z dimension of v_means3d, but will put that in a different branch and PR.


liruilong940607 (approved):
LGTM!

PR #123, opened 2024-02-08
Add camera pose and projection gradient flow

jh-surh:
- Added gradient backward flow to `project_gaussians_backward_kernel` for `viewmat` and `projmat` 
- Added a test script `examples/test_pose_grad.py` to test pose gradient update. The script interface is the same as [`simple_trainer.py`](https://github.com/jh-surh/gsplat/blob/main/examples/simple_trainer.py):
```sh
python examples/test_pose_grad.py
python examples/test_pose_grad.py --img_path path/to/image
```
- Validated on nerfstudio Splatfacto in this PR: https://github.com/nerfstudio-project/nerfstudio/pull/2885

### DIFF
--- a/examples/requirements.txt
+++ b/examples/requirements.txt
@@ -1,3 +1,4 @@
 numpy
 tyro
 Pillow
+pytorch3d @ git+https://github.com/facebookresearch/pytorch3d@v0.7.5
--- a/examples/test_pose_grad.py
+++ b/examples/test_pose_grad.py
@@ -82,7 +82,9 @@ class SimpleTrainer:
             convention="XYZ",
         )
         self.noisy_pose[:3, :3] = torch.matmul(self.noisy_pose[:3, :3], rot_noise)
-        self.noisy_pose[:3, 3] += (torch.rand(3, device=self.device) - 0.5) * bd  # +-1 units
+        self.noisy_pose[:3, 3] += (
+            torch.rand(3, device=self.device) - 0.5
+        ) * bd  # +-1 units
         self.pose_params = p3d.se3_log_map(
             self.noisy_pose.unsqueeze(0).permute(0, 2, 1)
         ).squeeze(0)
@@ -114,7 +116,15 @@ class SimpleTrainer:
         times = [0] * 3  # project, rasterize, backward
         for iter in range(iterations):
             start = time.time()
-            xys, depths, radii, conics, num_tiles_hit, cov3d = _ProjectGaussians.apply(
+            (
+                xys,
+                depths,
+                radii,
+                conics,
+                compensation,
+                num_tiles_hit,
+                cov3d,
+            ) = _ProjectGaussians.apply(
                 self.means,
                 self.scales,
                 1,
@@ -181,7 +191,15 @@ class SimpleTrainer:
         self.save_image(out_img, "objective")
 
         # save image prior to camera pose optimazation
-        xys, depths, radii, conics, num_tiles_hit, cov3d = _ProjectGaussians.apply(
+        (
+            xys,
+            depths,
+            radii,
+            conics,
+            compensation,
+            num_tiles_hit,
+            cov3d,
+        ) = _ProjectGaussians.apply(
             self.means,
             self.scales,
             1,
@@ -219,7 +237,15 @@ class SimpleTrainer:
         times = [0] * 3  # project, rasterize, backward
         for iter in range(iterations):
             start = time.time()
-            xys, depths, radii, conics, num_tiles_hit, cov3d = _ProjectGaussians.apply(
+            (
+                xys,
+                depths,
+                radii,
+                conics,
+                compensation,
+                num_tiles_hit,
+                cov3d,
+            ) = _ProjectGaussians.apply(
                 self.means,
                 self.scales,
                 1,
--- a/examples/test_pose_grad.py
+++ b/examples/test_pose_grad.py
@@ -1,0 +1,325 @@
+import math
+import os
+import time
+from pathlib import Path
+from typing import Optional
+
+import numpy as np
+import torch
+import tyro
+from gsplat.project_gaussians import _ProjectGaussians
+from gsplat.rasterize import _RasterizeGaussians
+from PIL import Image
+import pytorch3d.transforms as p3d
+from torch import Tensor, optim
+import torch.nn.functional as F
+
+
+class SimpleTrainer:
+    """Trains random gaussians to fit an image."""
+
+    def __init__(
+        self,
+        gt_image: Tensor,
+        num_points: int = 2000,
+    ):
+        self.device = torch.device("cuda:0")
+        self.gt_image = gt_image.to(device=self.device)
+        self.num_points = num_points
+
+        BLOCK_X, BLOCK_Y = 16, 16
+        fov_x = math.pi / 2.0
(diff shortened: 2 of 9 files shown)


ichsan2895:
`python examples/test_pose_grad.py --img_path path/to/image` needs pytorch3D. Maybe we add the requirement to `setup.py`?

jh-surh:
@ichsan2895,
Good idea, but I think it may be better to add it to examples/requirements.txt

examples/requirements.txt
```diff
@@ -1,3 +1,4 @@
 numpy
 tyro
-Pillow
\ No newline at end of file
+Pillow
+pytorch3d @ git+https://github.com/facebookresearch/pytorch3d@v0.7.5
```
kerrj, line 4: It seems like only the tests rely on p3d, it would be nice to remove as a dependency since it's a pretty big one and if we only need it for matrix transforms there's more common lightweight options like scipy.

kerrj:
It would be good to compare these camera grads to the camera grads from the torch implementation with autograd, [this PR](https://github.com/nerfstudio-project/gsplat/pull/118) adds a torch implementation which should have correct (although slow) gradients since they're computed from torch.

oseiskar:
Also this would be simpler without the redundant `projmat` (see https://github.com/nerfstudio-project/gsplat/pull/97). For camera calibration optimization, you would very likely want to optimize fx, fy, cx, cy (and possibly other distortion parameters if support to such are added) and not the redudant OpenGL/NDC-style projection matrix.

jh-surh:
> Also this would be simpler without the redundant `projmat` (see #97). For camera calibration optimization, you would very likely want to optimize fx, fy, cx, cy (and possibly other distortion parameters if support to such are added) and not the redudant OpenGL/NDC-style projection matrix.

I'm not certain if removing the `projmat` is a good idea. I think it may add some nice functionality in choosing the type of projection you want during the gaussian projection step.
What do you guys think? @ichsan2895 @kerrj

examples/requirements.txt
jh-surh: Unfortunately I needed the matrix transforms to be differentiable so that the gradients can flow into the Euler angle and position parameters. I tried implementing my own transformations, but ran into some instability in training. I have been using `se3_exp_map` and `se3_log_map`, but if you have any replacements I am open to try them.

oseiskar:
> I'm not certain if removing the `projmat` is a good idea. I think it may add some nice functionality in choosing the type of projection you want during the gaussian projection step. What do you guys think? @ichsan2895 @kerrj

@jh-surh  I elaborated this here https://github.com/nerfstudio-project/gsplat/pull/97#issuecomment-1951381539. Everything that can be implemented with projection matrices can also be implemented by extending the "intrinsics" to support other camera models than the ideal pinhole (fx,fy,cx,cy). It's also relatively simple to convert any typical (non-ortho) OpenGL projection matrix to (fx,fy,cx,cy) format (and throw away the near & far clip terms which do not matter in this context)

jh-surh:
> > I'm not certain if removing the `projmat` is a good idea. I think it may add some nice functionality in choosing the type of projection you want during the gaussian projection step. What do you guys think? @ichsan2895 @kerrj
> 
> @jh-surh I elaborated this here [#97 (comment)](https://github.com/nerfstudio-project/gsplat/pull/97#issuecomment-1951381539). Everything that can be implemented with projection matrices can also be implemented by extending the "intrinsics" to support other camera models than the ideal pinhole (fx,fy,cx,cy). It's also relatively simple to convert any typical (non-ortho) OpenGL projection matrix to (fx,fy,cx,cy) format (and throw away the near & far clip terms which do not matter in this context)

Would it require work "extending the intrinsics to support other camera models" in terms of calculating gradients? If so, it seems like something we can do in a separate PR since the current nerfstudio implementation uses projmat for its own projection matrix.

> It would be good to compare these camera grads to the camera grads from the torch implementation with autograd, [this PR](https://github.com/nerfstudio-project/gsplat/pull/118) adds a torch implementation which should have correct (although slow) gradients since they're computed from torch.

Thank you for your suggestion! I found a bug trying to compare the gradient w.r.t. the torch implementation. I have updated the test script and have passed the project gaussian test. Seems like the current main branch is failing 3 of the test scripts though. Someone should look at that.

oseiskar:
> Would it require work "extending the intrinsics to support other camera models" in terms of calculating gradients? If so, it seems like something we can do in a separate PR since the current nerfstudio implementation uses projmat for its own projection matrix.

Yes, and I agree that such stuff should definitely be implemented in some other PR.

However, the projection matrix is fed to gsplat is _not_ needed by Nersftudio for any other purpose, as demonstrated here https://github.com/SpectacularAI/nerfstudio/commit/bd234898725b1df46984a8f8b0925e2de43d4581 (works fine with #97).

My argument is that the more stuff is built on top of the current API with separate and redundant "projmat" + fx,fy,cx,cy, the more difficult it becomes to simplify it. I now remembered/realized the situation with the redundant API is worse than I described earlier: "projmat" is actually not a "projection matrix" in any standard sense but a model-view-projection matrix, adding another layer of confusion).

This complications caused by this API are clearly visible in this PR: why you need to compute the "projection matrix" gradient for pose optimization in the first place is because the parameter known as projmat is set to `projmat @ viewmat` in Nerfstudio, and this term depends on `viewmat`, which is the thing you actually want to modify in pose optimization.

Without the `@ viewmat` part, projection matrix gradients would only be needed for camera instrinsics optimization, but for this purpose, you would _also_ need to add gradients w.r.t. fx, fy, cx, cy to work with the current API. Without "projmat", you would only need the latter.

jh-surh:
> My argument is that the more stuff is built on top of the current API with separate and redundant "projmat" + fx,fy,cx,cy, the more difficult it becomes to simplify it. I now remembered/realized the situation with the redundant API is worse than I described earlier: "projmat" is actually not a "projection matrix" in any standard sense but a model-view-projection matrix, adding another layer of confusion).
> 
> This complications caused by this API are clearly visible in this PR: why you need to compute the "projection matrix" gradient for pose optimization in the first place is because the parameter known as projmat is set to `projmat @ viewmat` in Nerfstudio, and this term depends on `viewmat`, which is the thing you actually want to modify in pose optimization.
> 
> Without the `@ viewmat` part, projection matrix gradients would only be needed for camera instrinsics optimization, but for this purpose, you would _also_ need to add gradients w.r.t. fx, fy, cx, cy to work with the current API. Without "projmat", you would only need the latter.

I gotta say this makes a lot of sense. Now that I think of it, this would cause the gradient for `projmat` to flow to `viewmat`, which I'm not sure is the right move.

kerrj:
I'm thinking of merging #127 because of how much simpler it is and it seems to give good results, @jh-surh thoughts?

maturk:
> I'm thinking of merging #127 because of how much simpler it is and it seems to give good results, @jh-surh thoughts?

I agree. Both are equivalent to the best of my understanding.

jh-surh:
@kerrj @maturk @oseiskar I left a comment on https://github.com/nerfstudio-project/gsplat/pull/127.
TL;DR large difference between pure pytorch gradient and approximated `v_viewmat`.

kerrj:
Closing this since the 1.0 version includes more exact gradients I believe

PR #125, opened 2024-02-09
Fix core tests

jb-ye:
The core tests are broken since https://github.com/nerfstudio-project/gsplat/pull/118 Because the `test_project_gaussians_backward` was not properly annotated.


vye16 (approved):
Thank you, LGTM!

PR #126, opened 2024-02-09
fixed missing return type in simple_trainer.py

Brunwo:
Hi, I found that the example from the readme does not work without this change 
simple_trainer.py and project_gaussians.py are inconsistent at the moment


jb-ye:
Thanks for the fix! Nice catch.

kerrj (approved):
thanks!

Issue #68, opened 2023-11-02
nd rasterizer is 10x slower than rasterizer


zubair-irshad:
Hi, Great work! nd rasterizer is around 10x slower than sh rasterizer. To be precise, my model inference time with sh rasterization is 0.008s which gives me >100FPS as described in the original gaussian splatting paper but just adding nd rasterizaiton reduces it to 0.075 s and 13 FPS.

Is there a way to make it better? Any intuition would be greatly appreciated. With nd rasterization, it looks like we lose the benefits i.e. speed of gaussian splatting. Thank you again for the awesome work!

vye16:
Hi! What N are you using? In rasterization, each pixel requires an N-d
array of workspace memory. For RGB, we can fit that in register memory, and
can specify this statically at compile time. We wrote N-d for the case that
the necessary workspace exceeds available register memory, and must be in
global memory. This means we can't make the same kinds of optimizations in
the RGB rasterizer. If this is the case for you, then you can either stick
with the global memory situation, or you can rasterize in batches with the
current optimized RGB rasterizer (channels 0-3, 3-6, etc). We're
considering adding an in-between version of the rasterizer for
`MAX_REGISTER_CHANNELS=16` with similar optimizations to the RGB rasterizer.

zubair-irshad:
Thank you for the great intuition and detailed response. My channel size is currently 29 but I am considering increasing the feature size to 128 or even 256 which my worry is it will be slower than 13 FPS. I will try the batched RGB rasterizer as you suggested in a for-loop manner and see if it gives a higher FPS, thank you!

zubair-irshad:
@vye16 Reporting back what I found. Implementing a for loop to rasterize multiple channels in batches i.e. 0-3, 3-5 etc instead of ND rasterization is slightly worse in performance and I didn't find it to improve performance. My guess is due to the for loop which has to run 10 times for the channel size I am trying i.e. 30. Any other intuition to improve performance is greatly appreciated, thank you!

Just to provide more specifics, per iteration time for 640 by 480 image for nd rasterization with N=30 is `~74-76ms`, with batched (the one I shared above is `82-85ms`, with just sh i.e. 3 channel rendering it is `16ms`. The same results translate to fps numbers during inference i.e. `13fps` for batched and nd_rasterization with N=30 vs `>100fps` for sh rasterization only

zubair-irshad:
[...]

PR #129, opened 2024-02-13
Refactor interfaces to hide tile_bounds and allow dynamic block_size

kerrj:
Previously the user had to work with black magic `tile_bounds` and the values of `BLOCK_X`, `BLOCK_Y` were hardcoded in CUDA so only `16` was a valid input. Now, CUDA has been refactored to take any value of block size <=16, and the tile_bounds computations have been completely hidden from the user. The new interface for `project_gaussians` and `rasterize_gaussians` instead accept an integer `block_size` input, from which tile bounds is calculated from.

There are 2 goals for this change: 
1. Clean up the interface to users for tile_bounds and get rid of hardcoded BLOCK_X in CUDA, and
2. Dynamic block size will allow more sophisticated shared memory handling down the road, allowing for smaller block sizes to rasterize higher dimensional values more efficiently. This will address speed issues in ND rasterization like seen in #68 when implemented.


kerrj:
I tested this PR in nerfstudio with block sizes 2-16, training works for all sizes tested

vye16 (changes requested):
this looks good to me, tested on block sizes 2^n from 1 to 4. please rename `block_size` to `block_width` so we can distinguish from actual block size (total size) vs side length. otherwise lgtm I'll approve after those changes. make sure to make the accompanying changes in nerfstudio

vye16 (approved):
great lgtm

maturk:
@kerrj, can you explain the need for this in relation to shared memory. I thought that regardless of the num of threads/workers in a CUDA block, the shared memory size is fixed? So changing the block_width, and hence thread count, here would not make a difference to the allocatable shared memory. Let me know if I have any misunderstandings.

kerrj:
There are some nuances with shared memory size; 1) Some GPUs allow allocating more than 48KB of shared memory if you dynamically allocate it 2) launching a kernel with too much shared memory requested limits how many blocks can launch at the same time, which can starve the processor, in which case launching with a smaller block size is actually faster since more of the processors can be utilized. 3) `regardless of the num of threads/workers in a CUDA block, the shared memory size is fixed` This is true about the maximum shared memory, however you can launch with less to free up shared memory for other blocks executing.

PR #136, opened 2024-02-27
Fix backprop grad of cov2d / scales and unit tests

jb-ye:
I found big bug! The calculation of cov2d's grad is incorrect and the unit test doesn't report errors because the test cases are too weak. In this PR, I basically fix the test cases to be proper as well as (1) cov2d's grad calculation from v_conics (2) scales' grad should be affected by glob_scale. I also bump the version because the change is really necessary for splatfacto.


gsplat/_torch_impl.py
```diff
@@ -133,11 +133,16 @@ def quat_to_rotmat(quat: Tensor) -> Tensor:
...
+    assert quat.shape[-1] == 4, quat.shape
+    return normalized_quat_to_rotmat(F.normalize(quat, dim=-1))
+
+
 def scale_rot_to_cov3d(scale: Tensor, glob_scale: float, quat: Tensor) -> Tensor:
     assert scale.shape[-1] == 3, scale.shape
     assert quat.shape[-1] == 4, quat.shape
     assert scale.shape[:-1] == quat.shape[:-1], (scale.shape, quat.shape)
-    R = quat_to_rotmat(quat)  # (..., 3, 3)
+    R = normalized_quat_to_rotmat(quat)  # (..., 3, 3)
```
jb-ye, line 145: The cuda code currently doesn't implement the backprop gradient of normalize(...).

PR #139, opened 2024-03-01
Fix v_mean3d in project_g and v_conic calculation in *rasterize_backward

jb-ye:
Previous https://github.com/nerfstudio-project/gsplat/pull/136 doesn't fully solve the problem and in fact caused the splatfacto performance degradation. This is because there are two bugs that canceling each other in training GS

- the first bug which is solved in https://github.com/nerfstudio-project/gsplat/pull/136 is missing a 0.5 multiplier when computing v_cov2d from v_conics.y
- the second bug which is solved in this PR is to remove an extra 0.5 multiplier when computing v_conics.y in rasterize_backward call

Effectively they cancel out each other to some degree in splafacto when updating cov2d. 

By only fixing the first bug but not the second, the performance of splatfacto significantly degraded. In an [example dataset](https://drive.google.com/file/d/1pUOWW6w1OOzIkAwji_jkgKOugTinN8VO/view?usp=sharing) ichsan2895 provided me, PSNR degrades from 25.97 to 25.40. Post this PR, the PSNR performance is restored.

There is a third bug in updating v_mean3d which @vye16 mentioned to me earlier. This change also fix the bug and no side effect is observed (either improvement or regression).


kerrj (approved):
lgtm!

PR #140, opened 2024-03-05
add backprop grad of opacity compensation

jb-ye:
# Why

In the previous PR https://github.com/nerfstudio-project/gsplat/pull/117 I found some side efforts when training with a lot tiny gaussians: the gaussian splatting training is less likely to densify tiny gaussians and as a result a lot fine details cannot be constructed when training in antialiased mode. 

See an example below: the first image is the classic mode, where even though the color of bag at zoom in is not correct (due to aliasing), the geometric textures are reconstructed. The second image is the anti-aliased mode, while the color at zoom in is correct, but a lot geometric textures are lost. 
classic
antialiased_v1

After some deep dives, I realized that the root cause is we didn't back-propagate gradients through the opacity compensation in previous PR and as a result the gradient calculated for cov2d of tiny splats are no longer close to the true gradient when the compensation factor is away from 1. Adding back the backprop grad of opacity compensation seems to fix the issue in my experiments. Below is a video demo of this PR.

https://github.com/nerfstudio-project/gsplat/assets/132313008/44782424-3aee-4375-92fa-e35be7149587

# Sanity Checks

I further did some sanity checks and found that have the true gradient is beneficial in marginally improving PSNR metrics as well. So I consider this to be a necessary step toward enabling antialiased mode in splatfacto by default. 

PSNR metrics of bicycle dataset (downsampled by factor of 4) in mipnerf360
* classic 25.25 
* classic (coarse-to-fine) 25.39
* antialiased (coarse-to-fine) 25.46

Besides the GS metrics, I also modify the current unit test of project_gaussians to ensure the implementation of gradient is numerical correct.


kerrj (approved):
lgtm

ichsan2895:
Maybe we should upgrade/bump version `gsplat` to 0.1.8 and force nerfstudio repo for not using 0.1.7 since it broken by #136

PR #144, opened 2024-03-13
Small CUDA changes for nd rasterization

kerrj:


gsplat/cuda/csrc/forward.cu
```diff
@@ -269,15 +265,14 @@ __global__ void nd_rasterize_forward(
             }
 
             const float next_T = T * (1.f - alpha);
-            if (next_T <= 1e-4f) { // this pixel is done
-                // we want to render the last gaussian that contributes and note
-                // that here idx > range.x so we don't underflow
+            if (next_T <= 1e-4f) {
```
vye16, line 268: why remove the comments?

gsplat/cuda/csrc/backward.cu
```diff
@@ -79,15 +79,11 @@ __global__ void nd_rasterize_backward_kernel(
         const float sigma =
             0.5f * (conic.x * delta.x * delta.x + conic.z * delta.y * delta.y) +
             conic.y * delta.x * delta.y;
-        if (sigma < 0.f) {
-            valid = 0;
-        }
+        valid &= (sigma >= 0.f);
```
vye16, line 82: out of curiosity is this for a reason?

vye16 (approved):
minor changes that look reasonable to me, aside from removing some helpful comments

gsplat/cuda/csrc/backward.cu
kerrj: More thread coalescence, since they can all execute one & operation rather than branching with an if statement.

gsplat/cuda/csrc/forward.cu
kerrj: TBH I didn't really understand what the comment was saying so I removed it, but I stuck it back if you think it's helpful

PR #149, opened 2024-03-26
remove projmat from API

kerrj:
I think it's cleaner to just entirely remove the argument from the API right now since users will eventually have to update this.


vye16 (changes requested):
looks good, can you also update `examples/simple_trainer.py` ty

kerrj:
updated

vye16 (approved):
lgtm

PR #152, opened 2024-03-30
Normalize quaternions in PyTorch Implementation

Mxbonn:
I was using the PyTorch implementation for educational purposes and noticed the implementation did not quite match the cuda one.
In the cuda implementation the quaternions are always normalized but not in the PyTorch reference implementation.


jb-ye:
Hi @Mxbonn Thanks for point out the discrepancy between cuda and pytorch. 

I made the change from `quat_to_rotmat` to `normalized_quat_to_rotmat` in a previous PR because the back-propagate gradient implementation in cuda doesn't account the normalization step. (I did notice that in cuda implementation, the forward call always normalize quaternion, but the background call currently doesn't implement the proper grad for normalization op. )

In order to test the correctness the gradient implementation of cuda using `test/test_project_gaussians.py`, I had to use `normalized_quat_to_rotmat` in pytorch implementation such that the grad test can pass. 

BTW, in splatfacto, we always [normalize quaternions](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/models/splatfacto.py#L715) before triggering the project_gaussians() call.

Issue #153, opened 2024-03-30
Known issue: quaternion normalization behaviors are inconsistent between forward and backward call. 


jb-ye:
I found this issue earlier, but i think it might be good to document here. 

Current cuda implementation of project_gaussians()

the forward call normalize quaternion (See [here](https://github.com/nerfstudio-project/gsplat/blob/main/gsplat/cuda/csrc/forward.cu#L500) and [here](https://github.com/nerfstudio-project/gsplat/blob/main/gsplat/cuda/csrc/helpers.cuh#L151-L157)).

the backward call normalize the input quats but does not calculate the grad w.r.t the normalize step. See details here: https://github.com/nerfstudio-project/gsplat/blob/main/gsplat/cuda/csrc/backward.cu#L467

jb-ye:
@vye16 @kerrj Just want to let you this issue. It would not trigger any issue as long as the user use normalized quats as input.

PR #155, opened 2024-03-31
Check for normalized quats.

Mxbonn:
From https://github.com/nerfstudio-project/gsplat/pull/152#issuecomment-2028434952 it seems like the quaternions should always be normalized because the back-propagate gradient implementation in cuda doesn't account the normalization step. 
This PR adds a check for that.
Tagging @jb-ye .

### DIFF
--- a/gsplat/project_gaussians.py
+++ b/gsplat/project_gaussians.py
@@ -34,7 +34,7 @@ def project_gaussians(
        means3d (Tensor): xyzs of gaussians.
        scales (Tensor): scales of the gaussians.
        glob_scale (float): A global scaling factor applied to the scene.
-       quats (Tensor): rotations in quaternion [w,x,y,z] format.
+       quats (Tensor): rotations in normalized quaternion [w,x,y,z] format.
        viewmat (Tensor): view matrix for rendering.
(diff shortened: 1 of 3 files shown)


gsplat/project_gaussians.py
```diff
@@ -57,6 +57,7 @@ def project_gaussians(
         - **cov3d** (Tensor): 3D covariances.
     """
     assert block_width > 1 and block_width <= 16, "block_width must be between 2 and 16"
+    assert (quats.norm(dim=-1) - 1 < 1e-6).all(), "quats must be normalized"
```
jb-ye, line 60: It looks good to me. But I want @vye16 to verify the change. See #153 for the reason of why we want to merge this PR.

brentyi: Likely minor, but are there performance concerns here? This seems ~equivalent to a `torch.cuda.synchronize()` call.

jb-ye: I think this is fine.

brentyi: ok!

jb-ye (approved):
prepare for next version 0.1.10

PR #156, opened 2024-04-01
normalize the quats in torch implements

gezhixuan:
When the quaternion is not normalized, project_gaussians in _torch_impl.py  and forward.cu give different results.I think the reason for this problem is that the latter has an additional normalization operation for quaternions compared to the former.

https://github.com/nerfstudio-project/gsplat/blob/2135488e27f2ee3c00de13d16b583605050614f0/gsplat/cuda/csrc/helpers.cuh#L150-L153


jb-ye:
See #155, #152 and #153

PR #164, opened 2024-04-16
create JIT compilation configs if necessary

ychfan:
`_get_build_directory` needs additional permission to access file system, only call it if necessary


liruilong940607 (approved):
Thanks for the fix! This makes sense to me!

PR #166, opened 2024-04-23
Implement AbsGS

liruilong940607:
Abs grad for gaussian splitting criterion. See
- "AbsGS: Recovering Fine Details for 3D Gaussian Splatting"
- "EfficientGS: Streamlining Gaussian Splatting for Large-Scale High-Resolution Scene Representation"


liruilong940607:
> * AbsGS: Recovering Fine Details for 3D Gaussian Splatting

@niujinshuchong  It is really nice that you bring this up! Sorry that we missed it but actually your paper is already under my radar and is exactly the next paper I plan to read in depth (I have skimmed through it and very like it!). 

That being said, I just realized there is a subtle but probably important difference between the AbsGS paper and yours. It seems like what's proposed in the AbsGS  paper is `abs-sum-norm` (first abs grad_x and grad_y, then sum over all pixels, and  take the norm after than). What's in your paper seems to be `abs-norm-sum` (please correct me if I'm wrong).

I was implementing following the AbsGS paper but since you bring this up I actually now feels your way might makes more sense. I wonder what's your thoughts on this.

+cc @jb-ye

niujinshuchong:
Hi, thanks for the kind words.

We implemented it as `abs-sum-sum`, please check [here](https://github.com/autonomousvision/mip-splatting/blob/main/submodules/diff-gaussian-rasterization/cuda_rasterizer/backward.cu#L594-L595), the l2-norm is a typo in the paper. I agree different norms could be used. But as a result, different thresholds should be chosen.

Another difference of our implementation to the other papers is that we do not disable 3DGS's original metric but instead combine them together. We first select the candidates for densification with 3DGS's metric and check how many percentage of the Gaussians are selected and then select the same number of candidates with our metric. By doing so, we don't need to tune the hyperparameter for the threshold and it is guarantee that we will get at least the same quality as the original 3DGS's method.

jb-ye:
The original approach (sum-norm) and the new approach (abs-sum-norm) is like two extremes. What really matters here to identify how to split the gaussians into two such that variance of pixel partial gradients within each group are minimized. 

Intuitively, the ideal approach is very like the one used in training a regression tree, each time we find the best split to minimize some form of errors / variances, and we do so by growing the data structure (e.g. tree) in a greedy way.

gsplat/rasterize.py
```diff
@@ -231,6 +233,11 @@ def backward(ctx, v_out_img, v_out_alpha=None):
                 v_out_alpha,
             )
 
+        # Abs grad for gaussian splitting criterion. See
+        # - "AbsGS: Recovering Fine Details for 3D Gaussian Splatting"
+        # - "EfficientGS: Streamlining Gaussian Splatting for Large-Scale High-Resolution Scene Representation"
+        xys.absgrad = v_xy_abs
```
kerrj, line 239: I wonder if this will cause problems in splat training? currently there's a `xys.retain_grad()` in the nerfstudio implementation that's needed to keep the `.grad` property around. We should verify that this absgrad property works in the nerfstudio training pipeline

kerrj:
I tested this in nerfstudio by switching `self.xys.grad` to `self.xys.absgrad` and for some reason much fewer gaussians get split and the quality is worse. (screenshots attached, in particular look at the trees in the background)
Normal grads:
Screenshot 2024-04-25 at 12 29 36 PM
this PR (`xys.absgrad`)
Screenshot 2024-04-25 at 12 29 31 PM

liruilong940607:
Yeah from my limited experiments I also didn't find it to be a life saver. It might also because that there are many thresholds entangled together, making it hard to predict how a single one would affect the overall procedure

jb-ye:
One thing one should keep an eye on is the large gaussian culling. This is the major difference of splatfacto and the original GS code. You can try to disable it and see if things improve with AbsGrad.

jb-ye:
@liruilong940607 Hey, I found a bug in your implementation. See my latest commit.

gsplat/rasterize.py
jb-ye: The only issue of this hack is it doesn't do accumulation like xys.grad if multiple backward() calls are triggered. Let's say if one rasterize gaussians twice in the forward call, the backward would be called twice, but only the latest v_xy_abs is saved. Maybe add a comment here to indicate the difference in how backward call would be updating xys.grad and xys.absgrad.

jb-ye (approved):
I have experimented with this PR in splatfacto, the implementation stays correct.

PR #168, opened 2024-04-26
remove some magic numbers

YertleTurtleGit:


jb-ye:
I intended to keep these magic numbers as they corresponds exactly to the [reference code](https://jcgt.org/published/0002/02/06/) from the original authors, so people can verify the correctness/consistency of the implementation. But I think a better way here is instead of replacing these magic numbers, add some comments like:

```
result[..., 0] = 0.2820947917738781 #  SH_C0
```

What do you think?

YertleTurtleGit:
> I intended to keep these magic numbers as they corresponds exactly to the [reference code](https://jcgt.org/published/0002/02/06/) from the original authors, so people can verify the correctness/consistency of the implementation. But I think a better way here is instead of replacing these magic numbers, add some comments like:
> 
> ```
> result[..., 0] = 0.2820947917738781 #  SH_C0
> ```
> 
> What do you think?

Hey, thanks for the early comment. Didn't expect that (:
And regarding that: Makes sense. I'll handle that in the coming days if you don't mind.
Cheers.

PR #170, opened 2024-04-28
Add backward gradient for background

KevinXu02:
Add backward gradient for background
(related with https://github.com/nerfstudio-project/nerfstudio/pull/3100)

### DIFF
--- a/gsplat/rasterize.py
+++ b/gsplat/rasterize.py
@@ -225,12 +225,7 @@ class _RasterizeGaussians(Function):
                 v_out_img,
                 v_out_alpha,
             )
-        # print(v_out_img)
-        # print("final_ts shape", final_Ts.shape)
-        # v_out_img = torch.tensor(v_out_img)
         v_background = torch.matmul(v_out_img.float().view(-1, 3).t(), final_Ts.float().view(-1, 1)).squeeze()
-
-        # v_background = torch.mm(v_out_img.view(-1, 3).t, final_Ts.view(-1, 1)).squeeze() # type: ignore
         return (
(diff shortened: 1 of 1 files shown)


gsplat/rasterize.py
```diff
@@ -117,10 +115,7 @@ def forward(
         num_intersects, cum_tiles_hit = compute_cumulative_intersects(num_tiles_hit)
 
         if num_intersects < 1:
-            out_img = (
-                torch.ones(img_height, img_width, colors.shape[-1], device=xys.device)
-                * background
-            )
+            out_img = torch.ones(img_height, img_width, colors.shape[-1], device=xys.device) * background
```
jb-ye, line 118: Should we change the annotation of 
`background: Optional[Float[Tensor, "channels"]] = None` to `background: Float[Tensor, "channels"]`. The implementation seems will rather if background is None.

gsplat/rasterize.py
```diff
@@ -230,7 +225,8 @@ def backward(ctx, v_out_img, v_out_alpha=None):
                 v_out_img,
                 v_out_alpha,
             )
-
+        v_background = torch.matmul(v_out_img.float().view(-1, 3).t(), final_Ts.float().view(-1, 1)).squeeze()
```
jb-ye, line 228: Should we have a check that (Because I believe v_background could be some overhead at rendering time)
```
v_background = None
if background. requires_grad:
    v_background = ...
```

KevinXu02: thanks for the advice. fixed

jb-ye (approved):

PR #171, opened 2024-04-29
return result array for lower sh degrees

YertleTurtleGit:
Not sure about this one, but shouldn't the result array be returned even for lower sh degrees?


jb-ye (approved):
Oh, yeah, thx for the catch!

PR #172, opened 2024-04-29
CUDA V2

liruilong940607:
I dramatically rewrite the CUDA part to make the code runs more efficient, with cleaner code structure, and some extra functionalities.

New features:
- **Support batch rendering in one pass:** N cameras -> images with shape (N, H, W, C). This allows for sampling evenly distributed cameras during every training iteration, thus could potentially lead to better convergence. With this, we probably don't need gradient accumulation anymore. Also, for rendering GS, if some latency can be allowed, we can use batched rendering to further boost FPS.
- **Supports fully PyTorch Auto-Grad:** Provides a version of the rasterization & projection implementation that the gradients are fully managed by PyTorch Auto-Grad (see `_rendering` in `experimental/cuda/__init__.py`). This allows for easier try-out for the new ideas as the modifications can probably done purely on the python side. It could also serves as a gradient checker for the CUDA implementation.
- **Fused implementation with verified Gradients:** Every gradient in this PR is carefully checked against PyTorch Auto-Grad. On the other hand, the current GSplat implementation has flaws (errors) in the gradients including
-- 1. FoV clip is applied in the [forward pass](https://github.com/nerfstudio-project/gsplat/blob/2ddee2a7d62d079dc2e169b5aca86c795db9753e/gsplat/cuda/csrc/forward.cu#L448-L471) but not the [backward pass](https://github.com/nerfstudio-project/gsplat/blob/2ddee2a7d62d079dc2e169b5aca86c795db9753e/gsplat/cuda/csrc/backward.cu#L413-L422). As a result, the gradients are incorrect on some GSs.
-- 2. The backward suffers from a numerical issue caused by [using the remained transmittance output from the forward pass](https://github.com/nerfstudio-project/gsplat/blob/2ddee2a7d62d079dc2e169b5aca86c795db9753e/gsplat/cuda/csrc/backward.cu#L179) because the remained transmittance can be very small (~1e-4). Double precision instead of float should be used for it.
- **Faster FWD and BWD (esp. in ND case):**  See table below for details.

Gradient Diff results again PyTorch Auto Grad (code in `experimental/test.py`):

|    | viewmats | scales | quats | means |
| -------- | ------- | ------- | ------- | ------- |
| GSplat |  42.8834    | 2e-4 | 5e-4 | 4e-3 |
| This PR | 4e-3  | 2e-7 | 5e-6 | 3e-6 |

Runtime Performance (in ms) for Forward (code in `experimental/profile.py`):

|  Forward, Channel=1  | Batch=1 | Batch=8 | Batch=64 | 
| -------- | ------- | ------- | ------- | 
| GSplat | 1.492 | 9.921 |  76.295 |
| This PR  | 1.146 | 3.593 | 28.948 |

|  Forward, Channel=32  | Batch=1 | Batch=8 | Batch=64 | 
| -------- | ------- | ------- | ------- | 
| GSplat | 5.078 | 32.824 | 264.146 |
| This PR | 1.930 | 7.773 | 64.461 |

Runtime Performance for Backward (disable viewmats gradients):

|  Backward, Channel=1  | Batch=1 | Batch=8 | Batch=64 | 
| -------- | ------- | ------- | ------- | 
| GSplat | 1.594 | 10.600 | 84.395 |
| This PR | 1.562 | 7.409 | 60.402 |

|  Backward, Channel=32  | Batch=1 | Batch=8 | Batch=64 | 
| -------- | ------- | ------- | ------- | 
| GSplat | 10.242 | 80.935 | 653.627 |
| This PR | 7.850 | 43.392 | 353.733 |

Runtime Performance for Backward (enable viewmats gradients):

|  Backward, Channel=3  | Batch=1 | Batch=8 | Batch=64 | 
| -------- | ------- | ------- | ------- | 
| GSplat | 2.721 | 16.629 | 130.977 |
| This PR | 2.183 | 10.697 | 85.817 |


liruilong940607:
It is not fully compatible with the master yet, including:
 - anti-aliasing compensation term is not yet implemented.
 - AbsGrad is to yet implemented.
 - Fused background is not yet implemented.
 
 But they should be easy to add and should not affect performance too much.

liruilong940607:
Borrowing the splatfacto model from nerfstudio into `experimental/example.py`, I was able to do a quick test on how this PR affects splatfacto on the case of single batch size, RGB rendering:

- Training time stays the same (I'm not even sure if gsplat is the main time cost in the training. Probably not)
- Evaluation on 24 of the image in the Garden scene is 35% faster (0.2185s -> 0.1619s). This indicates the viewer FPS can be better.
- Evaluation results are also marginally better: LPIPS 0.15073 -> 0.14789

gsplat/experimental/cuda/_warpper.py
```diff
@@ -0,0 +1,588 @@
+from typing import Callable, Tuple
```
fbcotter, line 1: nit: rename file to `gsplat/experimental/cuda/_wrapper.py`?

gsplat/experimental/cuda/_torch_impl.py
```diff
@@ -0,0 +1,290 @@
+import struct
+from typing import Tuple
+
+import torch
+from torch import Tensor
+
+
+def _quat_scale_to_covar_perci(
```
fbcotter, line 8: why `perci` over `preci` or `precis`?

gsplat/experimental/example.py
```diff
@@ -0,0 +1,1188 @@
+# ruff: noqa: E741
```
fbcotter, line 1: nice! Thanks for adding this file

gsplat/experimental/cuda/__init__.py
```diff
@@ -0,0 +1,146 @@
...
+        width,
+        height,
+        tile_size,
+        isect_offsets,
+        gauss_ids,
+    )
+    return render_colors, render_alphas
+
+
+def _rendering(means, quats, scales, opacities, colors, viewmats, Ks, width, height):
```
fbcotter, line 53: is the `_` naming convention in this file meant to follow what you have in `_torch_impl` where all the native pytorch methods are led with underscores and the cuda methods aren't? It looks like this function might be meaning to test out the pytorch methods which are imported but not called

fbcotter:
Thanks! Looking forward to trying out some faster rendering 🚀 . I'm surprised about the no difference to splatfacto training times, I can do some profiling on the models I'm training but I thought it was a large part of the time used. Perhaps a naive question - wouldn't we need to have a `.get_outputs_batch` method where we feed in batches of camera poses and intrinsics to really see a speedup?

gsplat/experimental/cuda/_warpper.py
liruilong940607: oh great catch!

gsplat/experimental/cuda/_torch_impl.py
liruilong940607: thanks for catching it yeah its a typo should be preci

gsplat/experimental/cuda/__init__.py
liruilong940607: I need to cleanup `__init__.py` and define these functions somewhere else and rename them properly. For this `_rendering` function, the current usage of it is to test against the `rasterize_to_pixels` CUDA implementation only.   So functions other than that part are still using the CUDA implementation to minimize the numerical difference for the test. But I do plan to make all other functions in it using torch impl functions in the end.

gsplat/experimental/cuda/_warpper.py
```diff
@@ -0,0 +1,588 @@
...
+        - Covariance matrices. If `triu` is True the returned shape is [N, 6], otherwise [N, 3, 3].
+        - Precision matrices. If `triu` is True the returned shape is [N, 6], otherwise [N, 3, 3].
+    """
+    assert quats.dim() == 2 and quats.size(1) == 4, quats.size()
+    assert scales.dim() == 2 and scales.size(1) == 3, scales.size()
+    quats = quats.contiguous()
+    scales = scales.contiguous()
+    covars, percis = _QuatScaleToCovarPerci.apply(
+        quats, scales, compute_covar, compute_perci, triu
+    )
```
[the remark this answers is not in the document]
liruilong940607: From what I remember, this implementation is 1 to 2 magnitude of faster than pure torch impl. I was also using it to check my CUDA impl.

liruilong940607:
> Thanks! Looking forward to trying out some faster rendering 🚀 . I'm surprised about the no difference to splatfacto training times, I can do some profiling on the models I'm training but I thought it was a large part of the time used. Perhaps a naive question - wouldn't we need to have a `.get_outputs_batch` method where we feed in batches of camera poses and intrinsics to really see a speedup?

Yeah if you train with batches then I would expect a noticeable speedup. The current example that I took from nerfstudio does not trivially support batch though. More upgrade is needed on that.

hturki:
This is very exciting @liruilong940607! One quick question - am I correct in assuming that the batch rendering supports multiple cameras against the same scene (ie: against the same set of gaussians), and not a potentially different set of gaussians per camera? I can think of some hacky ways of doing the latter (ie: move sets of gaussians very far apart), if it's it not a huge lift would be nice to have less hacky API support.

liruilong940607:
> This is very exciting @liruilong940607! One quick question - am I correct in assuming that the batch rendering supports multiple cameras against the same scene (ie: against the same set of gaussians), and not a potentially different set of gaussians per camera? I can think of some hacky ways of doing the latter (ie: move sets of gaussians very far apart), if it's it not a huge lift would be nice to have less hacky API support.

Yeah currently it is designed for the same scene. To support multiple scenes, I guess we would have to assume each scene contains different number of gaussians, which means GSs can not be trivially batched like cameras. This would make it a little bit more painful to implement than support batched cameras but not impossible!

hturki:
Makes sense - FWIW I think it would be reasonable to expect that each scene has the same number of gaussians (worse-case one could find a way of "zero-padding" gaussians). It would still be better than the status quo.

fbcotter:
> > This is very exciting @liruilong940607! One quick question - am I correct in assuming that the batch rendering supports multiple cameras against the same scene (ie: against the same set of gaussians), and not a potentially different set of gaussians per camera? I can think of some hacky ways of doing the latter (ie: move sets of gaussians very far apart), if it's it not a huge lift would be nice to have less hacky API support.
> 
> Yeah currently it is designed for the same scene. To support multiple scenes, I guess we would have to assume each scene contains different number of gaussians, which means GSs can not be trivially batched like cameras. This would make it a little bit more painful to implement than support batched cameras but not impossible! Though it probably is not in our near future roadmap.

What about supporting the same scene (or at least the same number of gaussians) but with possibly moved gaussian centres? E.g. if the means were allowed to move through time? Currently the `projection` function only allows unbatched means and covariances as far as I can tell

liruilong940607:
Closing this PR in favor of #183

PR #182, opened 2024-05-26
Improve numerical stability of _torch_impl.project_cov3d_ewa

oseiskar:
Fixes the a minimal set of numerical issues in `_torch_impl` so that it can be used in actual training and not just unit tests. This makes testing and optimizing various things a lot easier (example [here](https://github.com/SpectacularAI/gsplat/commit/f40a3eeec4967c210674c21ec7ebb15d3ea1ac76)). Similar to one of the goals in https://github.com/nerfstudio-project/gsplat/pull/172. However, #172 seems to be still a work-in-progress, so this PR still makes sense too and should not conflict with #172.

The numerical issues were mostly related to:
 1. Computing things for Gaussians that are behind the current camera
 2. Trying to differentiate $\sqrt x$ too close to $x = 0$

### DIFF
--- a/gsplat/_torch_impl.py
+++ b/gsplat/_torch_impl.py
@@ -6,7 +6,7 @@ import torch
 import torch.nn.functional as F
 from jaxtyping import Float
 from torch import Tensor
-from typing import Tuple, Literal
+from typing import Tuple, Literal, Optional
 
 
 def compute_sh_color(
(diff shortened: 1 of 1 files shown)


oseiskar:
> Looks good to me. Have you tested the test_project_gaussians.py ?

Yes. Still passing. (Just double checked after realizing the passing CI build does not apparently test CUDA things)

jb-ye (approved):
looks good to me

PR #183, opened 2024-05-26
V 1.0 Major Update

liruilong940607:
This is a continue development based on #172. There are many major features in this update, including:

- Comparing to the [official implementation](https://github.com/graphdeco-inria/gaussian-splatting), gsplat enables up to **4x less training memory footprint**, and up to **2x less training time** on Mip-NeRF 360 captures, and potential more on larger scenes. 

- Support extremely large scene rendering, which is magnitudes faster than the official CUDA backend [diff-gaussian-rasterization](https://github.com/graphdeco-inria/diff-gaussian-rasterization). 

- Extra features, including batch rasterization, N-D feature rendering (faster), depth rendering, sparse gradient etc.


gsplat/__init__.py
```diff
@@ -1,166 +1,113 @@
...
-    compute_cov2d_bounds,
-    get_tile_bin_edges,
-)
-from .sh import spherical_harmonics
-from .version import __version__
 import warnings
 
+from .rendering import rasterization
+from ._helper import rasterization_legacy_wrapper, rasterization_inria_wrapper
+from .version import __version__
```
maturk, line 5: the version should probably be bumped from 0.1.11 to 1.0.0

tests/test_basic.py
maturk, line 1: I have a few tests failing on my machine:

image

maturk (commented):
Fantastic stuff, I left some small comments but will look at this more in detail.

jb-ye:
Let's have a one version update in 0.1.xx to check in those untagged commits before we bump into 1.0

docs/source/migration/migration_inria.rst
```diff
@@ -0,0 +1,27 @@
+Migratite from `diff-gaussian-rasterization <https://github.com/graphdeco-inria/diff-gaussian-rasterization>`_
```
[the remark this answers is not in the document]
liruilong940607: done

gsplat/__init__.py
liruilong940607: done

tests/test_basic.py
liruilong940607: I wonder if you can reproduces it. I also experienced some numerical instability on my end.

liruilong940607:
Ok I think I've cleanup all the TODOs in my plates. I'm a bit afraid of merging it into the main with such a big update. Anyone wants to mess around with it or proof read the doc?  @jb-ye @maturk @brentyi @jake-austin @vye16  

Here is the preview version of the doc as for the latest commit: https://6656ab91e5ef673fdf2219e3--inquisitive-mandazi-513b7e.netlify.app/

I'll pin the current main to a 0.1.x version before merging this PR.

liruilong940607:
Ok seems like I missed the cleanup on .md files, and also needs to fix the testing is somehow failing in github actions

jb-ye:
> With that, I think I'm fine with not blocking the v1.0 release by these extra supports

Sounds good to me.

liruilong940607:
Ok fixed the tests and update the md files. So good to go now.

liruilong940607:
Just pinned main to `0.1.12`

gsplat/cuda/_wrapper.py
```diff
@@ -0,0 +1,1120 @@
...
+    assert viewmats.size() == (C, 4, 4), viewmats.size()
+    means = means.contiguous()
+    covars = covars.contiguous()
+    viewmats = viewmats.contiguous()
+    return _WorldToCam.apply(means, covars, viewmats)
+
+
+def projection(
+    means: Tensor,  # [N, 3]
+    covars: Tensor,  # [N, 6] or None
```
hturki, line 143: (noted this when I was porting https://github.com/nerfstudio-project/gsplat/compare/cuda_v2...hturki:gsplat:ht/cuda_v2_batch?expand=1 to this branch)

I would suggest annotating the parameters that might be None as Optional[Tensor]

gsplat/cuda/csrc/rasterization.cu
```diff
@@ -0,0 +1,1245 @@
...
+    }
+    const int64_t cid_enc = cid << (32 + tile_n_bits);
+
+    int64_t depth_id_enc = (int64_t) * (int32_t *)&(depths[idx]);
+    int64_t cur_idx = (idx == 0) ? 0 : cum_tiles_per_gauss[idx - 1];
+    for (int i = tile_min.y; i < tile_max.y; ++i) {
+        for (int j = tile_min.x; j < tile_max.x; ++j) {
+            int64_t tile_id = i * tile_width + j;
+            // e.g. tile_n_bits = 22:
+            // camera id (10 bits) | tile id (22 bits) | depth (32 bits)
```
vye16, line 79: nice!

gsplat/cuda/_wrapper.py
```diff
@@ -0,0 +1,1120 @@
...
+    assert means.size() == (N, 3), means.size()
+    assert covars.size() == (N, 3, 3), covars.size()
+    assert viewmats.size() == (C, 4, 4), viewmats.size()
+    means = means.contiguous()
+    covars = covars.contiguous()
+    viewmats = viewmats.contiguous()
+    return _WorldToCam.apply(means, covars, viewmats)
+
+
+def projection(
```
vye16, line 141: projection = quat_scale_to_covar -> world_to_cam -> persp_proj
maybe rename to something like full_gaussian_projection so it's not too similar with persp_proj or something?

gsplat/cuda/_wrapper.py
```diff
@@ -0,0 +1,1120 @@
...
+        if not ctx.needs_input_grad[0]:
+            v_means = None
+        if not ctx.needs_input_grad[1]:
+            v_covars = None
+        if not ctx.needs_input_grad[2]:
+            v_viewmats = None
+        return v_means, v_covars, v_viewmats
+
+
+class _Projection(torch.autograd.Function):
```
vye16, line 666: same comment about name

gsplat/cuda/csrc/rasterization.cu
```diff
@@ -0,0 +1,1245 @@
...
+    CHECK_INPUT(depths);
+    if (rindices.has_value()) {
+        CHECK_INPUT(rindices.value());
+    }
+    if (cindices.has_value()) {
+        CHECK_INPUT(cindices.value());
+    }
+    bool packed = means2d.dim() == 2;
+
+    int N, nnz, totel_elems;
```
vye16, line 107: nit: totel_elems -> total_elems

gsplat/cuda/_wrapper.py
liruilong940607: oh yes good catch! I think was generally following this rule but I clearly missed this one!

gsplat/cuda/csrc/rasterization.cu
```diff
@@ -0,0 +1,1245 @@
...
+        offsets.fill_(0);
+    }
+    return offsets;
+}
+
+/****************************************************************************
+ * Rasterization
+ ****************************************************************************/
+
+__global__ void rasterize_to_indices_iter_kernel(
```
vye16, line 254: okay so this function only rasterizes a range of indices determined by step0 to step1
maybe rename to rasterize_to_indices_in_range_kernel and rename step0 and step1 to range_start and range_end respectively?

gsplat/cuda/csrc/rasterization.cu
```diff
@@ -0,0 +1,1245 @@
...
+            image_width, image_height, tile_size, tile_width, tile_height,
+            tile_offsets.data_ptr<int32_t>(), gauss_ids.data_ptr<int32_t>(),
+            transmittances.data_ptr<float>(), chunk_starts.data_ptr<int32_t>(), nullptr,
+            out_gauss_ids.data_ptr<int32_t>(), out_pixel_ids.data_ptr<int32_t>());
+    }
+    return std::make_tuple(out_gauss_ids, out_pixel_ids);
+}
+
+template <uint32_t COLOR_DIM>
+__global__ void rasterize_to_pixels_fwd_kernel(
```
vye16, line 478: really nice, also parallelize of cameras with multidim thread block 👍 👍

gsplat/cuda/csrc/rasterization.cu
```diff
@@ -0,0 +1,1245 @@
...
+    torch::Tensor renders = torch::empty({C, image_height, image_width, channels},
+                                         means2d.options().dtype(torch::kFloat32));
+    torch::Tensor alphas = torch::empty({C, image_height, image_width, 1},
+                                        means2d.options().dtype(torch::kFloat32));
+    torch::Tensor last_ids = torch::empty({C, image_height, image_width},
+                                          means2d.options().dtype(torch::kInt32));
+
+    at::cuda::CUDAStream stream = at::cuda::getCurrentCUDAStream();
+
+    // TODO: an optimization can be done by passing the actual number of channels into
```
vye16, line 678: good idea, though not necessary for this version unless you're motivated XD

gsplat/cuda/_wrapper.py
```diff
@@ -0,0 +1,1120 @@
...
+          of {`means`, `covars`, `quats`, `scales`} will be a sparse Tensor in COO layout. Default: False.
+        calc_compensations: If True, a view-dependent opacity compensation factor will be computed, which
+          is useful for anti-aliasing. Default: False.
+
+    Returns:
+        A tuple:
+
+        If `packed` is True:
+
+        - **rindices**. The row indices of the projected Gaussians. Int32 tensor of shape [nnz].
```
vye16, line 201: I think this needs to be renamed to be clearer, row/column being camera and gaussian idcs right?

vye16 (changes requested):
amazing ruilong, I learned so much! I just left a couple of comments about naming and interfaces (talking more offline) and we can merge this in super soon!

Linusnie:
Hi, this looks like a great update! I did some testing for pose gradients and noticed a pretty significant slowdown (roughly 3x). It's probably not noticeable in terms of overall reconstruction time, although might be worth looking into.

There is a workaround by transforming the gaussians to camera space and letting autodiff handle the pose jacobians (as is done in SplaTAM https://github.com/spla-tam/SplaTAM/issues/64). However, this adds some overhead as you need an extra matrix-to-quaternion conversion for rotating the covariance matrices.

My timings experiments can be found here https://github.com/Linusnie/render-splats/blob/main/timings.ipynb

liruilong940607:
> Hi, this looks like a great update! I did some testing for pose gradients and noticed a pretty significant slowdown (roughly 3x). It's probably not noticeable in terms of overall reconstruction time, although might be worth looking into.
> 
> There is a workaround by transforming the gaussians to camera space and letting autodiff handle the pose jacobians (as is done in SplaTAM [spla-tam/SplaTAM#64](https://github.com/spla-tam/SplaTAM/issues/64)). This variant is almost as fast as the legacy pose gradients, but adds some overhead as you need an extra matrix-to-quaternion conversion for rotating the covariance matrices.
> 
> My timings experiments can be found here https://github.com/Linusnie/render-splats/blob/main/timings.ipynb

Hey @Linusnie thanks for this profiling!! It's a very interesting experiment.

First of all, I would like to state that the camera pose gradients are flawed in of `gsplat v.0.1.X`, as the `spherical_harmonics` is differentiable w.r.t view directions, but [the gradients are not implemented](https://github.com/nerfstudio-project/gsplat/blob/main/gsplat/sh.py#L97-L104) (essentially putting a `detach()` there).

It is implemented in this PR to make sure the camera pose received the **full** gradients.

So the slow down you observe should be mainly coming from there, i.e., the d (SH) / d(viewdirs) is currently quite a bottleneck. And by transforming Gaussians into the camera coordinate system (your third experiment), you would actually also need to transform the SH coeffs to the camera coordinate system using the camera pose, which is missing in your currently implementation. So as a result, your third experiment skips the gradient calculation of d (SH) / d(viewdirs), resulting a significant speedup.

But this is an interesting thought! The camera gradients in this PR is numerically correct but we didn't put too much efforts on optimizing it for speed. This idea might be able to give us something if we could figure out how SH would be handled.

Linusnie:
@liruilong940607 I see, thanks for the explanation! Very nice to have the full gradient implemented. I do transform the view directions in the legacy version with this line `view_dirs = bmv(gs_data.rotation.T, torch.nn.functional.normalize(gs_data.means - camera_center))`, which I guess should be more or less equivalent to transforming the SH coefficient (maybe with some difference due to truncation error).

liruilong940607:
@Linusnie Ah that is smart! But I guess the view directions are not transformed in your third experiment (V1.0 transformed Gaussians)? As this would need to be injected into `rasterization` function in V1.0.

Linusnie:
@liruilong940607 that's right. Although I just realized the view directions already depend on the pose in "V1.0 transformed" via the gaussian means (which are transformed by the pose):

```python
camera_center = torch.zeros(3, device='cuda', requires_grad=False)
means = (camera_pose @ gs_data).means
view_dirs = torch.nn.functional.normalize(means - camera_center)
colors = gsplat.cuda._wrapper.spherical_harmonics(gs_data.max_sh_degree, view_dirs, gs_data.features)
torch.autograd.grad(colors.mean(), camera_pose)
# returns nonzero value
```

So looks like the cause would be something else after all. Maybe some overhead due to supporting batched poses?

liruilong940607:
@Linusnie I guess what I was trying to express is that only transforming `view_dirs` is not enough -- i.e., it will end up a different output `colors` from this function (because you are using a camera space view direction to query the SH basis instead of a world space view direction)

```bash
colors = gsplat.cuda._wrapper.spherical_harmonics(gs_data.max_sh_degree, view_dirs, gs_data.features)
```

That being said, the SH coeffs (`gs_data.features`) is actually also coordinate-system-dependent, which should also be "transformed" by the camera pose, so that the above function will give the same output between camera space inputs and world space inputs. But I'm not sure how the SH coeffs can be transformed by camera pose.

Linusnie:
@liruilong940607 I think there should be no need to transform `gs_data.features` directly. For the legacy variant I keep track of the rotation that has been applied to the gaussians via `gs_data.rotation`. This rotation can then be used to transform view directions from camera coordinates to the original SH coordinates. In both cases you actually end up passing the same viewdirs to `spherical_harmonics`: `m - (-R^Tt) = R^T(Rm + t)` (LHS: regular variant, RHS: transformed variant, `m` is the gaussian mean, `R,t` is the pose). So in terms of rendered colors they are equivalent

As you said this rotation would need to be passed as an additional argument to `rasterization` to keep the output correct in V1.0. With my previous comment I just wanted to point out that gradients are propagated through the view directions both with or without this correction, via the gaussian means

liruilong940607:
@Linusnie Ah you are right! I didn't see it through that this transformation leads to the same `viewdirs` as before! But this still wouldn't avoid the computation of d (colors) / d (view_dirs) in `spherical_harmonics`, which I guess meaning it wouldn't speed up things?

```
view_dirs = torch.nn.functional.normalize(means - camera_center) # should additionally be rotated by camera_pose[:3, :3] to be geometrically consistent
colors = gsplat.cuda._wrapper.spherical_harmonics(gs_data.max_sh_degree, view_dirs, gs_data.features)
```

Linusnie:
@liruilong940607 that's the thing, the slow execution for pose gradients seems to rather be connected to whether `viewmats` has requires_grad=True or not. Not on whether gradients are propagated through the spherical harmonics. Based on that `render_splats(camera_pose, gs_data)` is much slower than `render_splats(eye(4), camera_pose @ gs_data)`, even though d(SH)/d(viewdirs) is computed in both cases

Linusnie:
@liruilong940607 I've narrowed it down to this line here https://github.com/nerfstudio-project/gsplat/blob/v1.0/gsplat/rendering.py#L274! Looks like the backwards pass through the re-indexing of `camtoworlds` is what's taking so much time (even though `rindicies` is all zeros). Probably because I'm working with a small example scene (the lego excavator) and one camera.

Using replacing the second part with `camtoworlds[:, :3, 3].repeat((len(cindices), 1))` resolves the issue. This doesn't handle the multi-camera case, but maybe something similar could be made to work.

liruilong940607:
Very interesting! Does that change bring the timing to the same between the two options?

gsplat/cuda/csrc/rasterization.cu
liruilong940607: I'll leave this to next PR ... lol

gsplat/cuda/_wrapper.py
liruilong940607: good point! I rename it to `fully_fused_projection`

gsplat/cuda/_wrapper.py
liruilong940607: done! `rindices` -> `camera_ids`, `cindices` -> `gaussian_ids`

gsplat/cuda/csrc/rasterization.cu
liruilong940607: good point. done!

vye16 (approved):
Nice update on radixsort, 👏 👏

CITATION.cff
maturk, line 1: Do we update the citation along with this PR or reserve that for later? @liruilong940607 , @vye16

liruilong940607: any reason that we want to reserve it for later?

maturk (approved):

CITATION.cff
liruilong940607: ok communicated offline!

gsplat/rendering.py
```diff
@@ -0,0 +1,562 @@
...
+        camtoworlds = torch.inverse(viewmats)
+        if packed:
+            dirs = means[gaussian_ids, :] - camtoworlds[camera_ids, :3, 3]
+        else:
+            dirs = means[None, :, :] - camtoworlds[:, None, :3, 3]
+        colors = spherical_harmonics(
+            sh_degree, dirs, colors, masks=radii > 0
+        )  # [nnz, D] or [C, N, 3]
+        # make it apple-to-apple with Inria's CUDA Backend.
+        colors = torch.clamp_min(colors + 0.5, 0.0)
```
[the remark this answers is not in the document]
liruilong940607: Ah good point! that's a bug we should fix on this nerfstudio side!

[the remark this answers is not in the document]
liruilong940607: Fixed in https://github.com/nerfstudio-project/nerfstudio/pull/3207

PR #195, opened 2024-06-02
Fixed broken link in README.md

ajean757:
Link to evaluation page was broken, replaced with latest results


liruilong940607 (approved):
Thanks for the fix!

PR #200, opened 2024-06-05
2DGS implementation

FantasticOven2:
This is an implementation of 2DGS that tend to integrate it with gsplat library. 
Done:

- [x] Ray Splat intersect 

TODO:
- [ ] Depth distortion loss
- [ ] Normal loss
- [ ] nd_rasterization


maturk:
Hi @FantasticOven2, great start looks promising! A few things to start off, can you rebase your repo with the most recent main from gsplat and also not override the already existing forward/backward functions (but instead rename them as 3dgs and 2dgs respectively). e.g. `project_gaussians`->`project_gaussians_3dgs` and add your `project_gaussians_2dgs` method, similarly for rasterization and the cuda backend.

hardikdava:
Looking forward to integration!

PR #216, opened 2024-06-11
Fix wrong description in rasterization API doc

yifanlu0227:
Fix the tensor shape:
**render_colors**: The rendered colors. ~~[C, width, height, X]~~ [C, height, width, X] ...
**render_alphas**: The rendered alphas. ~~[C, width, height, 1]~~ [C, height, width, 1] ...

Also update the toy example to show the difference between width and height.

### DIFF
--- a/gsplat/rendering.py
+++ b/gsplat/rendering.py
@@ -148,12 +148,12 @@ def rasterization(
     Returns:
         A tuple:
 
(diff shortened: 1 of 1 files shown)


liruilong940607 (approved):
LGTM! thanks for the fix!

PR #217, opened 2024-06-11
workaround for n-dim rendering with n>32

wuzirui:
slice the input colors & background in 32-dim slices and iteratively rasterize to pixels

### DIFF
--- a/gsplat/rendering.py
+++ b/gsplat/rendering.py
@@ -295,20 +295,52 @@ def rasterization(
         colors = depths[..., None]
     else:  # RGB
         pass
-    render_colors, render_alphas = rasterize_to_pixels(
-        means2d,
-        conics,
-        colors,
-        opacities,
-        width,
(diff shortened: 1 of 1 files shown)


liruilong940607:
oh yeah this is a nice work around. I do have plane to update something in the cuda kernel to allow for a direct processing on more channels but it's not yet pop onto the top 1 position of my todo list.

We can use this work around for now that unlocks users and lead the cuda tricks for future as performance improvement.

liruilong940607 (changes requested):
Would you mind also update the func doc together within this file? Basically removing the line of "up to 32 at the moment"

liruilong940607:
Format checking is failing. You could run
```
pip install .[dev]
black . gsplat/ tests/ examples/ profiling/ 
```

liruilong940607 (approved):
Thanks! LGTM!

PR #221, opened 2024-06-12
Support Batch Rasterization For Multiple GS Scenes

liruilong940607:
After this PR, we will  have the full batching support

The API of `rasterization()` will support passing in GS attributes with shape `[B, N, ...]`, in which `B` is the batch dimension for scenes, and `N` is the number of GS in each scene. For cameras, it will support cameras extrinsics with shape `[B, C, 4, 4]` and intrinsics with shape`[B, C, 3, 3]`, in which C is the number of cameras **per scene**. The rendered image would be with shape `[B, C, H, W, 3]`.

This API would support various batching use cases:
- Single scene multi-view (C) rendering. GS would be with shape `[1, N, ...]` and cameras would be with shape `[1, C, ...]`. The output would be with shape `[1, C, H, W, 3]`.
- Multiple scenes (B), each scene has ONE distinct viewpoint. GS would be with shape `[B, N, ...]` and cameras would be with shape `[B, 1, ...]`. The output would be with shape `[B, 1, H, W, 3]`.
- Multiple scenes (B), each scene has multiple (C) cameras. GS would be with shape `[B, N, ...]` and cameras would be with shape `[B, C, ...]`. The output would be with shape `[B, C, H, W, 3]`.

### DIFF
--- a/examples/simple_viewer.py
+++ b/examples/simple_viewer.py
@@ -56,7 +56,7 @@ if args.ckpt is None:
     N = len(means)
     print("Number of Gaussians:", N)
 
-    # batched render
+    # batched render on 2 scenes, each scene has C cameras.
     render_colors, render_alphas, meta = rasterization(
         means.expand(2, -1, -1),  # [B, N, 3]
         quats.expand(2, -1, -1),  # [B, N, 4]
@@ -69,6 +69,7 @@ if args.ckpt is None:
         height,
         render_mode="RGB+D",
     )
+    # [B, C, H, W, D]
     assert render_colors.shape == (2, C, height, width, 4)
     assert render_alphas.shape == (2, C, height, width, 1)
 
--- a/examples/simple_viewer.py
+++ b/examples/simple_viewer.py
@@ -17,6 +17,7 @@ import numpy as np
 import torch
 import torch.nn.functional as F
 import viser
+
 from gsplat._helper import load_test_data
 from gsplat.rendering import rasterization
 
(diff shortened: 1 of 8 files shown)


liruilong940607:
The code is moderately tested. So it feels good to go but i would like to do more tests before merging it in. I'm putting this PR out just to unlock users that have such a need urgently.

An example of usage can be found in `examples/simple_viewer.py`, in which it performs batched rendering and dumps a image `render.png` to the disk.

liruilong940607:
This PR assumes the number of GSs across scenes are the same. In the cases this is not true, a workaround could be padding them to the same size with zero opacity.

This is not optimal but this keeps the API clean and easy to impl.

maturk:
I failed to run the example.

Steps to reproduce: 
```
python examples/simple_viewer.py --scene_grid 3  --ckpt examples/results/benchmark/baseline/bicycle/ckpts/ckpt_29999.pt
```

Error:
```
  File "/home/turkulm1/miniconda3/envs/gsplat/lib/python3.10/site-packages/nerfview/_renderer.py", line 144, in run
    rendered = self.viewer.render_fn(task.camera_state, img_wh)
  File "/home/turkulm1/miniconda3/envs/gsplat/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/home/turkulm1/git/gsplat/examples/simple_viewer.py", line 161, in viewer_render_fn
    render_colors, render_alphas, meta = rasterization_fn(
  File "/home/turkulm1/git/gsplat/gsplat/rendering.py", line 259, in rasterization
    opacities = opacities[batch_ids, gaussian_ids]  # [nnz]
IndexError: shape mismatch: indexing tensors could not be broadcast together with shapes [2], [0]
```

Possible cause:
error seems to come on this line [here](https://github.com/nerfstudio-project/gsplat/blob/2b0de894232d21e8963179a7bbbd315f27c52c9c/gsplat/rendering.py#L257)

liruilong940607:
Oh yeah seems like this PR still has some bugs -- I just realized it yesterday. I'll change it into a draft for now

liruilong940607:
I'm gonna deprecate this PR bc I think I have a better (more general) way of implementing this

LeaFendd:
I'm very interested in this work, will this feature be available in the next release? @liruilong940607

liruilong940607:
We do have plan to support it but unfortunately I don't have an ETA

Ma-Weijian:
> This PR assumes the number of GSs across scenes are the same. In the cases this is not true, a workaround could be padding them to the same size with zero opacity.
> 
> This is not optimal but this keeps the API clean and easy to impl.

I don't think this is not optimal since language models also have [PAD] tokens when the sentence does not reach the maximal length.

BTW, I'm really looking forward to this nice feature!

PR #238, opened 2024-06-25
Implement 3DGS-MCMC in gsplat.

jefequien:
I ported the code from [3dgs-mcmc](https://github.com/ubc-vision/3dgs-mcmc) into gsplat for learning purposes. Is this something you guys would want to integrate? If so, I can clean up the PR more.

Visualization of training from a random initialization and a 1M gaussian cap.

https://github.com/nerfstudio-project/gsplat/assets/6775703/69f49999-8da8-4565-8d2c-e754564346dd

## Metrics

Preliminary metrics using the `benchmark.sh` script. MCMC runs explicitly set the number of splats to be the number 3DGS ends with. 

| PSNR   |   3dgs_random |   3dgs_sfm |   mcmc_random |   mcmc_sfm |
|---------|---------------|------------|---------------|------------|
| counter |         27.74 |      28.99 |         29.05 |      29.39 |
| kitchen |         30.41 |      31.36 |         31.54 |      31.82 |
| bonsai  |         30.78 |      32.13 |         32.28 |      32.76 |
| room    |         30.24 |      31.29 |         31.85 |      32.24 |

| Number of GSs  | 3dgs_random   | 3dgs_sfm   | mcmc_random   | mcmc_sfm   |
|---------|---------------|------------|---------------|------------|
| counter | 1.02M         | 1.19M      | 1.19M         | 1.19M      |
| kitchen | 1.33M         | 1.79M      | 1.79M         | 1.79M      |
| bonsai  | 1.25M         | 1.22M      | 1.22M         | 1.22M      |
| room    | 1.41M         | 1.58M      | 1.58M         | 1.58M      |

Screenshot 2024-06-26 at 2 35 48 PM


liruilong940607:
Wow this is super cool! We would love to have this support.

Actually we are considering of abstracting out the densification strategies from the `simple_trainer.py` as there is a growing list of papers with different densification strategies including 3DGS-MCMC. But that could be a refactor for me to do later after this PR. 

For this PR, considering there are many changes needed to be done in the `simple_trainer.py`, maybe it makes sense to just create another script, e.g. `simple_trainer_mcmc.py`, without affecting the original script, which is designed to be matching with the official implementation with the default arguements. I could do the refactor and merging later on.

Also, before this PR get accepted, it would be nice to verify the implementation via reproducing the metrics in the paper, and list of the differences with the original paper if there is any. We have a `benchmark.sh` script that should be helpful to benchmark the mipnerf360 dataset. (It's fine if you dont have the resources to run all the scenes, but we should at least verify it on one of the captures)

jefequien:
Great! I'll make the `simple_trainer_mcmc.py` script and do my best to reproduce the metrics in the paper. I don't actually have enough vram to fully train the bicycle scene but I should be able to reproduce some of the smaller scenes.

examples/simple_trainer.py
```diff
@@ -150,22 +156,32 @@ def adjust_steps(self, factor: float):
...
     batch_size: int = 1,
     feature_dim: Optional[int] = None,
     device: str = "cuda",
 ) -> Tuple[torch.nn.ParameterDict, torch.optim.Optimizer]:
-    N = points.shape[0]
+    if init_type == "sfm":
+        points = torch.from_numpy(parser.points).float()
+        rgbs = torch.from_numpy(parser.points_rgb / 255.0).float()
+    elif init_type == "random":
+        points = scene_scale * (torch.rand((init_num_pts, 3)) * 3 * 2 - 3)
```
liruilong940607, line 175: What is the reason for the `* 3` here? Is it for the nerf synthetic data that mostly lives in the range of [-1.5, 1.5]? Also does it work with COLMAP captures? Maybe yes to most of the captures because we normalize the scene. But as the scale normalization was based on camera distribution, it is not guaranteed that the scene lives in [-1.0, 1.0]. So I think we should have a scaler here to control the initialized range, which can be an argument with default value as 3. In fact, I thought that `init_scale` is for this purpose, but its not. Should we use `init_scale` here instead of multiplying it to the `dist_avg`?

examples/simple_trainer.py
```diff
@@ -150,22 +156,32 @@ def adjust_steps(self, factor: float):
...
+        rgbs = torch.rand((init_num_pts, 3))
+    else:
+        raise ValueError("Please specify a correct init_type: sfm or random")
 
+    N = points.shape[0]
     # Initialize the GS size to be the average dist of the 3 nearest neighbors
     dist2_avg = (knn(points, 4)[:, 1:] ** 2).mean(dim=-1)  # [N,]
     dist_avg = torch.sqrt(dist2_avg)
-    scales = torch.log(dist_avg).unsqueeze(-1).repeat(1, 3)  # [N, 3]
+    scales = torch.log(dist_avg * init_scale).unsqueeze(-1).repeat(1, 3)  # [N, 3]
```
liruilong940607, line 184: what's the idea of scaling up distances between points?

examples/simple_trainer.py
```diff
@@ -378,15 +396,15 @@ def train(self):
         max_steps = cfg.max_steps
         init_step = 0
 
-        scheulers = [
+        schedulers = [
```
liruilong940607, line 399: nice catch!

gsplat/cuda/csrc/bindings.h
```diff
@@ -213,4 +213,11 @@ fully_fused_projection_packed_bwd_tensor(
     const torch::Tensor &v_depths,                      // [nnz]
     const torch::Tensor &v_conics,                      // [nnz, 3]
     const at::optional<torch::Tensor> &v_compensations, // [nnz] optional
-    const bool viewmats_requires_grad, const bool sparse_grad);
\ No newline at end of file
+    const bool viewmats_requires_grad, const bool sparse_grad);
+
+std::tuple<torch::Tensor, torch::Tensor> ComputeRelocationCUDA(
```
liruilong940607, line 218: Could you change the name of this to the same pattern with the existing functions? I know the pattern we have is not a standard pattern but let's just keep them consistent for now.

gsplat/relocation.py
```diff
@@ -0,0 +1,63 @@
+import math
+import torch
+from .cuda._wrapper import _make_lazy_cuda_func
+
+N_max = 51
```
liruilong940607, line 5: lets make the global variable with capital letters. Also `binoms`

gsplat/relocation.py
```diff
@@ -0,0 +1,63 @@
...
+binoms = torch.zeros((N_max, N_max)).float().cuda()
+for n in range(N_max):
+    for k in range(n + 1):
+        binoms[n, k] = math.comb(n, k)
+
+
+def compute_relocation_cuda(opacity_old, scale_old, N):
```
liruilong940607, line 12: Could you mark the typing of the inputs and the output, and write a couple lines of code docs to explain the  
- what this function is doing / used for
- reference
- input & outputs (shape and meaning)
- warning / notes if there is any

Also we could get rid of the suffix `_cuda`, and add some checking on the inputs to make sure they are:
- cuda tensor
- with the expected shape
- and call `.contiguous()` to make sure the memory layout is continuous.

gsplat/relocation.py
```diff
@@ -0,0 +1,63 @@
...
+    new_opacity, new_scale = _make_lazy_cuda_func("compute_relocation")(
+        opacity_old, scale_old, N.int(), binoms, N_max
+    )
+    return new_opacity, new_scale
+
+
+def sample_alives(probs, num, alive_indices=None):
```
liruilong940607, line 19: Docs for this one as well

gsplat/cuda/csrc/utils.cu
```diff
@@ -0,0 +1,51 @@
...
+    int P,
+    float* opacity_old,
+    float* scale_old,
+    int* N,
+    float* binoms,
+    int n_max,
+    float* opacity_new,
+    float* scale_new)
+{
+	int num_blocks = (P + 255) / 256;
```
liruilong940607, line 47: add checkings here just like other cuda functions would be nice

liruilong940607 (commented):
Nice work! I'm dumping some feedbacks. Mostly what we need are the docs of the functions, which might be already in your plan.

examples/simple_trainer.py
jefequien: `* 3` specifies the extent of the randomized initialized points. The paper recommends 3x the camera extent but also says there is not a big difference between 1x and 3x. 1x would make more sense for synthetic scenes. Created variable `init_extent`.

examples/simple_trainer.py
jefequien: `init_scale` is for specifying the size of the initialized splats. MCMC prefers to start with smaller, higher opacity splats than 3DGS.
https://github.com/graphdeco-inria/gaussian-splatting/compare/main...ubc-vision:3dgs-mcmc:main#diff-487784439059d635b474c389cb30ba79bfd9d5d116dfd49cb83f52806273b8ffR136

examples/simple_trainer.py
jefequien: Assume `scene_scale` ~= the furthest camera in the COLMAP capture and the camera poses are centered around zero, I think this would work.

devernay:
It would be nice to also have comparisons agains splatfacto and splatfacto-big. splatfacto has AbsGrad, which may resolve at least *some* of the issues that 3DGS-MCMC tackles.

liruilong940607:
> It would be nice to also have comparisons against splatfacto and splatfacto-big. splatfacto has AbsGrad, which may resolve at least _some_ of the issues that 3DGS-MCMC tackles.

splatfacto is using this repo under the hood, including absgrad. In fact if you run our `simple_trainer.py` with `--absgrad, --grow_grad2d 0.0008` it is exactly the splafacto default behavior.

liruilong940607:
@jefequien Nice benchmarking! Thanks for doing that! It seems to be working very well! Do you know what is happening with  the stump scene though? That one seems to have a bigger drop.

One thing that I didn't notice before, is that the numbers for 3DGS in the MCMC paper are constantly higher than either the numbers reported in the official paper of 3DGS, or what its code produces (as marked on [our eval page](https://docs.gsplat.studio/main/tests/eval.html)). I'm wondering what leads to that, which might also explains the slightly drop on the metrics between the MCMC implementation you have comparing to what's the MCMC paper reported?

jefequien:
Yeah! The overall trend seems correct but I don't know about the discrepancy yet. The paper says they use the same number of Gaussians for both methods for a fair comparison but they never specified what that number is for each scene. Maybe their 3DGS implementation ends with a different number than gsplat?

For the stump scene, I only have a GPU with 8G of VRAM so I can't train 3DGS on the bicycle, garden, or stump scenes. I had arbitrarily set the Gaussian cap to 2M. I'll remove those lines for clarity.

examples/simple_trainer_mcmc.py
```diff
@@ -644,17 +644,17 @@ def add_new_gs(self, cap_max):
...
     @torch.no_grad()
     def add_noise_to_gs(self, last_lr):
-        L = build_scaling_rotation(
-            torch.exp(self.splats["scales"]), self.splats["quats"]
-        )
-        actual_covariance = L @ L.transpose(1, 2)
+        opacities = torch.sigmoid(self.splats["opacities"])
+        scales = torch.exp(self.splats["scales"])
+        quats_normalized = F.normalize(self.splats["quats"], dim=-1)
+        actual_covariance = scale_rot_to_cov3d(scales, 1.0, quats_normalized)
```
liruilong940607, line 650: I think you can just use 
```
from gsplat import quat_scale_to_covar_preci
actual_covariance, _ = quat_scale_to_covar_preci(self.splats["quats"], scales, compute_covar=True, compute_perci=False, triu=False)
```

`quat_scale_to_covar_preci` is with fused CUDA implementation so should be faster. Also it does normalization on quats internally so there is no need to explicitly call `F.normalize`. We plan to deprecate the support of `cuda_legacy` at some point in the future so I would try to avoid relying on functions from that.

gsplat/cuda/csrc/bindings.h
jefequien: Done.

gsplat/relocation.py
jefequien: Removed function.

gsplat/relocation.py
jefequien: Added. Let me know if I should elaborate more.

gsplat/cuda/csrc/utils.cu
jefequien: Added `assert(P != 0);`

gsplat/cuda/csrc/relocation.cu
```diff
@@ -0,0 +1,67 @@
...
+        new_scales[idx * 3 + i] = coeff * old_scales[idx * 3 + i];
+}
+
+std::tuple<torch::Tensor, torch::Tensor> compute_relocation_tensor(
+    torch::Tensor &old_opacities,
+    torch::Tensor &old_scales,
+    torch::Tensor &ratios,
+    torch::Tensor &binoms,
+    const int n_max
+) {
```
liruilong940607, line 44: Adding `DEVICE_GUARD` is necessary when it is being used in a torch distributed training system. Also would be nice to apply `CHECK_INPUT` to the input tensors to make sure they are contiguous in memory, otherwise it could silently lead to incorrect results that is hard to debug. See reference:

https://github.com/nerfstudio-project/gsplat/blob/0ac847d6efde31dba8c289c67241321a3cf9821f/gsplat/cuda/csrc/projection.cu#L159-L161

gsplat/cuda/csrc/relocation.cu
```diff
@@ -0,0 +1,67 @@
...
+    const int P = old_opacities.size(0);
+    assert(P != 0);
+
+    torch::Tensor final_opacities =
+        torch::full({P}, 0, old_opacities.options().dtype(torch::kFloat32));
+    torch::Tensor final_scales =
+        torch::full({3 * P}, 0, old_scales.options().dtype(torch::kFloat32));
```
liruilong940607, line 51: For what reason the share here is {3 * P} instead of {P, 3} ?

gsplat/cuda/csrc/relocation.cu
```diff
@@ -0,0 +1,67 @@
...
+
+    int num_blocks = (P + 255) / 256;
+    dim3 block(256, 1, 1);
+    dim3 grid(num_blocks, 1, 1);
+    compute_relocation_kernel<<<grid, block>>>(
```
liruilong940607, line 56: Could you update these lines into something similar with? This allows 1/ streaming control, 2/ zero input, 3/ global control on the threads per block.

https://github.com/nerfstudio-project/gsplat/blob/0ac847d6efde31dba8c289c67241321a3cf9821f/gsplat/cuda/csrc/projection.cu#L181-L188

gsplat/relocation.py
```diff
@@ -0,0 +1,52 @@
...
+N_MAX = 51
+BINOMS = torch.zeros((N_MAX, N_MAX)).float().cuda()
+for n in range(N_MAX):
+    for k in range(n + 1):
+        BINOMS[n, k] = math.comb(n, k)
+BINOMS = BINOMS.contiguous()
+
+
+def compute_relocation(
+    old_opacities: Tensor,  # [N]
```
liruilong940607, line 15: I guess we don't need the `old_` prefix in the input naming.

gsplat/cuda/csrc/bindings.h
```diff
@@ -213,4 +213,13 @@ fully_fused_projection_packed_bwd_tensor(
     const torch::Tensor &v_depths,                      // [nnz]
     const torch::Tensor &v_conics,                      // [nnz, 3]
     const at::optional<torch::Tensor> &v_compensations, // [nnz] optional
-    const bool viewmats_requires_grad, const bool sparse_grad);
\ No newline at end of file
+    const bool viewmats_requires_grad, const bool sparse_grad);
+
+std::tuple<torch::Tensor, torch::Tensor>
+compute_relocation_tensor(
+    torch::Tensor& old_opacities,
```
liruilong940607, line 220: Also here, the `old_` prefix on the naming could go away.

liruilong940607 (commented):
Awesome clean up! Just a few minor comments then we are good to go!

gsplat/cuda/csrc/relocation.cu
```diff
@@ -42,14 +42,16 @@ std::tuple<torch::Tensor, torch::Tensor> compute_relocation_tensor(
     torch::Tensor &binoms,
     const int n_max
 ) {
+    DEVICE_GUARD(opacities);
+    CHECK_INPUT(opacities);
+    CHECK_INPUT(scales);
+    CHECK_INPUT(ratios);
+    CHECK_INPUT(binoms);
+    torch::Tensor new_opacities = torch::zeros_like(opacities);
+    torch::Tensor new_scales = torch::zeros_like(scales);
```
liruilong940607, line 51: These could be `torch::empty_like` as the following kernels would write values to it anyway. Using empty_like could save up one kernel launch for filling in zeros.

gsplat/cuda/csrc/relocation.cu
jefequien: Used empty_like.

liruilong940607 (approved):
Thank you for all the updates!! To me this PR is ready to merge. Anything you would like to further update?

jefequien:
Nope. Thanks for the reviews!

ichsan2895:
How to use this feature with `ns-train splatfacto`?

liruilong940607:
> How to use this feature with `ns-train splatfacto`?

We are doing a code refactoring in nerfstudio to support this.

ichsan2895:
Can wait before it integrates with `ns-train splatfacto`

I already test it with my custom dataset. It is significantly increasing the quality
image

ichsan2895:
Hello, I find an strange haze with drjohnson dataset. Here is the [detailed info](https://github.com/nerfstudio-project/gsplat/issues/252).

PR #240, opened 2024-06-27
updating background for different rendering modes [D, ED, RGB+D, RGB+…

vye16:
…ED] to be compatible for rasterization

make sure that if a background color is passed in, we update it in accordance to the rendering mode before passing into `rasterize_to_pixels`.


liruilong940607:
Ah right we should have that!

liruilong940607 (approved):

PR #244, opened 2024-06-29
distortion loss impl.

liruilong940607:
Implement distortion loss for 3DGS from [Mip-NeRF 360](https://arxiv.org/abs/2111.12077):
Screenshot 2024-06-29 at 10 32 53 AM

Note: Since the distortion loss is a per-pixel value, I implement it as a per-pixel map rendered from the `rasterization()` function, with the same resolution with the rendered image. This could be reduced into a scalar with a `.mean()` call on it for backward. Having the distortion loss as a map instead of a scalar could be useful for visualization and analysis.

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -439,7 +439,8 @@ def test_isect(test_data):
 
 @pytest.mark.skipif(not torch.cuda.is_available(), reason="No CUDA device")
 @pytest.mark.parametrize("channels", [3, 32, 128])
-def test_rasterize_to_pixels(test_data, channels: int):
+@pytest.mark.parametrize("distloss", [False, True])
+def test_rasterize_to_pixels(test_data, channels: int, distloss: bool):
     from gsplat.cuda._torch_impl import _rasterize_to_pixels
     from gsplat.cuda._wrapper import (
         fully_fused_projection,
@@ -487,8 +488,14 @@ def test_rasterize_to_pixels(test_data, channels: int):
     opacities.requires_grad = True
     backgrounds.requires_grad = True
 
+    if distloss:
+        colors = torch.cat([colors, depths[..., None]], dim=-1)
+        backgrounds = torch.cat(
+            [backgrounds, torch.zeros_like(backgrounds[..., :1])], dim=-1
+        )
+
     # forward
-    render_colors, render_alphas = rasterize_to_pixels(
+    render_colors, render_alphas, render_distloss = rasterize_to_pixels(
         means2d,
         conics,
         colors,
@@ -499,8 +506,9 @@ def test_rasterize_to_pixels(test_data, channels: int):
         isect_offsets,
         flatten_ids,
         backgrounds=backgrounds,
+        distloss=distloss,
     )
-    _render_colors, _render_alphas = _rasterize_to_pixels(
+    _render_colors, _render_alphas, _render_distloss = _rasterize_to_pixels(
         means2d,
         conics,
         colors,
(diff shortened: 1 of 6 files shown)


maturk:
@liruilong940607, I think there are some version issues with the Nerfacc library. I have Nerfacc Version: 0.5.3 installed (with `pip install nerfacc`) however, it seems that this is not compatible with the argument list in your test code. In particular I tracked down the commit and [this](https://github.com/nerfstudio-project/nerfacc/commit/ea07af8ef46cad0fd297c6364d60209d36240739) has not been added to a new release, so pip installing nerfacc results in a version where  there is no "indices" arguments to your implementations of inclusive/exclusive sums. 

The error a user gets: `TypeError: exclusive_sum() got an unexpected keyword argument 'indices'`. 

This can be fixed by installing from the git source but just a heads up...

tests/test_basic.py
```diff
@@ -511,18 +519,26 @@ def test_rasterize_to_pixels(test_data, channels: int):
...
         flatten_ids,
         backgrounds=backgrounds,
+        distloss=distloss,
     )
     torch.testing.assert_close(render_colors, _render_colors)
     torch.testing.assert_close(render_alphas, _render_alphas)
+    torch.testing.assert_close(render_distloss, _render_distloss)
 
     # backward
     v_render_colors = torch.randn_like(render_colors)
```
maturk, line 529: Not an issue, just a general question regarding the use of `torch.autograd.grad(outputs * v_outputs ,inputs)` in the gradient tests. Why do we scale outputs by these random `v_outputs` weights? Shouldn't `torch.autograd.grad(outputs, inputs)` suffice for gradient checking?

maturk: Makes sense, thanks. Here is a small visual of the dist loss as a heatmap whilst training: 
dist loss

liruilong940607: Very cool color theme! What theme are you using here?

The reason I haven't merge this PR is that I didn't find this distortion loss to be helpful on ANY of the scenes in MipNeRF 360, which is a bit frustrated and not expected. 

Let me know if you find anything otherwise!

liruilong940607: I'm guessing the reason that this distortion loss isn't useful might have something to do with the imperfect depth accumulation formulation that we are currently using. The depth formulation from [Gaussian Opacity Fields](https://niujinshuchong.github.io/gaussian-opacity-fields/) or [RaDe-GS](https://baowenz.github.io/radegs/) seems to be more accurate, which might be a prerequisite for distortion loss to work.

FantasticOven2: I tried this distortion loss with 2DGS and below are the results:
(with distortion loss)
IMG_0672

(without distortion loss)
IMG_0673

Seems like the improvement is very subtle and focus on details; and this is also the case for the original 2DGS implementation; I will try the L2 version in the 2DGS paper next

jefequien:
@FantasticOven2 I think the speckle artifact you have in your normal image is from overflowing uint8. You may need to clamp or normalize your normals.

tests/test_basic.py
jefequien: For those like me who were wondering what the awesome color scheme was, I think I tracked it down to `cmr.voltage`. https://cmasher.readthedocs.io/user/sequential/voltage.html

keunhong:
Hi Ruilong, do you have any plans to merge this? I think it would be helpful in some more underconstrained settings.

liruilong940607:
> Hi Ruilong, do you have any plans to merge this? I think it would be helpful in some more underconstrained settings.

I haven't got a case where I find it useful and that's why I wasn't quite motivated to spend further efforts on merging it. But I guess it's true that all the cases I tried are well constrained.

I could try to find some cycles to merge this but no guarantee on the timeline!

PR #247, opened 2024-06-30
Clamp `ratios` in MCMC's `compute_relocation`

yzslab:
Relative issue and PR:
* https://github.com/ubc-vision/3dgs-mcmc/issues/8
* https://github.com/ubc-vision/3dgs-mcmc/pull/9

The `max` value here is different from the PR above. Because according to [compute_relocation in cuda](https://github.com/nerfstudio-project/gsplat/blob/fc92cc5efa4584f3f49eda37cd516f72aa8aa20c/gsplat/cuda/csrc/relocation.cu#L18-L25), I think the max valid value is `N_max` rather than `N_max - 1`.

### DIFF
--- a/gsplat/relocation.py
+++ b/gsplat/relocation.py
@@ -38,6 +38,7 @@ def compute_relocation(
     assert ratios.shape == (N,), ratios.shape
     opacities = opacities.contiguous()
     scales = scales.contiguous()
+    ratios.clamp_(min=1, max=N_MAX)
     ratios = ratios.int().contiguous()
 
     new_opacities, new_scales = _make_lazy_cuda_func("compute_relocation")(


liruilong940607:
LGTM. cc @jefequien for review

jefequien:
We also have the `+1` here so shouldn't it be the same as the original codebase? At the same time, I think clamping to N_max is correct. Should they be clamping to N_max - 1? I could be wrong.
https://github.com/nerfstudio-project/gsplat/blob/main/examples/simple_trainer_mcmc.py#L570

liruilong940607 (approved):
Yes I agree with @jefequien the clamping to N_max in this PR should be correct. I'll merge it then. Thanks @yzslab and @jefequien!

PR #250, opened 2024-07-02
shim struct so that we can add extra args internally without breaking C++ API

fwilliams:


liruilong940607:
Thanks! After chatting with @vye16 offline we feel maybe having a struct as input to the function and also a struct as output would make the most sense for backward compatibility? Something like

```
# in the header file
struct PROJECTION_FWD_ARGS;
struct PROJECTION_FWD_RETURNS;

PROJECTION_FWD_RETURNS projection_fwd(PROJECTION_FWD_ARGS);
```

Then the user could just rely on this header file to call things.

fwilliams:
Closing this after discussion with @liruilong940607

Issue #239, opened 2024-06-26
benchmark.sh failed by out of bounds for unit64 error


MasahiroOgawa:
the error message is below
"""
$ bash benchmark.sh 
Running bicycle
Traceback (most recent call last):
  File "/mnt/home/mas/proj/study/computer_vision/gsplat/examples/simple_trainer.py", line 16, in <module>
    from datasets.colmap import Dataset, Parser
  File "/mnt/home/mas/proj/study/computer_vision/gsplat/examples/datasets/colmap.py", line 8, in <module>
    from pycolmap import SceneManager
  File "/mnt/home/mas/proj/study/computer_vision/gsplat/.venv/lib/python3.11/site-packages/pycolmap/__init__.py", line 4, in <module>
    from .scene_manager import SceneManager
  File "/mnt/home/mas/proj/study/computer_vision/gsplat/.venv/lib/python3.11/site-packages/pycolmap/scene_manager.py", line 21, in <module>
    class SceneManager:
  File "/mnt/home/mas/proj/study/computer_vision/gsplat/.venv/lib/python3.11/site-packages/pycolmap/scene_manager.py", line 22, in SceneManager
    INVALID_POINT3D = np.uint64(-1)
                      ^^^^^^^^^^^^^
OverflowError: Python integer -1 out of bounds for uint64
Traceback (most recent call last):
  File "/mnt/home/mas/proj/study/computer_vision/gsplat/examples/simple_trainer.py", line 16, in <module>
    from datasets.colmap import Dataset, Parser
  File "/mnt/home/mas/proj/study/computer_vision/gsplat/examples/datasets/colmap.py", line 8, in <module>
    from pycolmap import SceneManager
  File "/mnt/home/mas/proj/study/computer_vision/gsplat/.venv/lib/python3.11/site-packages/pycolmap/__init__.py", line 4, in <module>
    from .scene_manager import SceneManager
  File "/mnt/home/mas/proj/study/computer_vision/gsplat/.venv/lib/python3.11/site-packages/pycolmap/scene_manager.py", line 21, in <module>
    class SceneManager:
  File "/mnt/home/mas/proj/study/computer_vision/gsplat/.venv/lib/python3.11/site-packages/pycolmap/scene_manager.py", line 22, in SceneManager
    INVALID_POINT3D = np.uint64(-1)
                      ^^^^^^^^^^^^^
OverflowError: Python integer -1 out of bounds for uint64
Running bonsai
Traceback (most recent call last):
[...]

PR #251, opened 2024-07-02
add poetry files, and fix numpy error and no data error.

MasahiroOgawa:
# Objective
Fix [the issue](https://github.com/nerfstudio-project/gsplat/issues/239); fix numpy version and add download data script.
The user will not bother with the same issue with me.


Atticuszz:
nice pyproject.toml.😝

maturk:
Hi, I have a few thoughts. Maybe you can move the download dataset script into gsplat/examples/datasets subfolder. Also, I would prefer not to have a poetry.lock, this is too restrictive and just adds another thing that should be carefully checked and also maintained for compatibility for various different machines. Also, gsplat can be used as a dependency which does not require the installation of the various requirements in the examples/ directory. I think we should not have both a setup.py and a pyproject. The project should support various python versions (currently the lock is set at ^3.11 which is 3.11>=) and cuda/torch versions (currently locked at ^2.3.1+cu121 which requires quite a new torch version >=2.3.1 and toolkit respectively). Do you think just updating the requirements.txt with the correct numpy verison suffices for correct installation of the examples?

MasahiroOgawa:
Got it! I will remove poetry files and move the download script later.

maturk:
I removed the pyproject and added a generic download script that currently just downloads mipnerf360, but can be expanded for other benchmark datasets if needed.

examples/datasets/download_dataset.py
```diff
@@ -0,0 +1,78 @@
...
+import subprocess
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Literal
+
+import tyro
+
+# dataset names
+dataset_names = Literal[
+    "mipnerf",
```
liruilong940607, line 13: Would "mipnerf360" make more sense to call it?

liruilong940607 (approved):
LGTM!

MasahiroOgawa:
Thanks! I read the code. Good download script!

PR #255, opened 2024-07-03
Separate C source into many translation units

fwilliams:
Enables faster compilation and easier code readability


liruilong940607 (commented):
LGTM!

vye16 (approved):
Hey! I support splitting these up into separate files, it'll be much easier to extend in the future. I think it would also be nice to add some header files, maybe for fully_fused_projection and rasterize_to_pixels, but that can be in the future, since we don't currently have headers for those right now. Otherwise this looks great to me!

yklcs:
Note that the default parallelism of Ninja uses num_cores + 2 parallel jobs which easily results in swapping/OoM.
I needed to set the `MAX_JOBS` environment variable to something more sane to get gsplat to compile with this PR merged.

liruilong940607:
> Note that the default parallelism of Ninja uses num_cores + 2 parallel jobs which easily results in swapping/OoM. I needed to set the `MAX_JOBS` environment variable to something more sane to get gsplat to compile with this PR merged.

Ah true. Do you know an elegant way to limit it in `setup.py` and  JIT compiling at here?
https://github.com/nerfstudio-project/gsplat/blob/2f0bb12f614eb28a5dea2a5422809bc9c388fdb2/gsplat/cuda/_backend.py#L67-L75

Ideally we should set a default parallelism to a reasonable one (e.g., 12) but still allow user to overwrite that with `MAX_JOBS`. 

cc @fwilliams

PR #259, opened 2024-07-04
template everything but don't actually support multiple precisions yet

fwilliams:
Don't merge before merging #256


liruilong940607 (approved):
LGTM!

PR #267, opened 2024-07-06
float3 -> vec3<float>; float2 -> vec2<float>;

liruilong940607:
To embrace half in the future, and because there is no `half3` in native CUDA, we here migrate from `float3` to `glm::vec3` which has been templated for `half`.

Speed wise I didn't see a big change.

Before:
Screenshot 2024-07-06 at 9 25 58 AM

After:
Screenshot 2024-07-06 at 9 25 50 AM


liruilong940607:
cc @fwilliams

PR #270, opened 2024-07-07
Start adding support for double, float16, and bfloat16

fwilliams:


fwilliams:
@liruilong940607

kerrj:
cool! will this have support for heterogenous types? eg represent means in float32 and colors/scales in float16? It seems like float16 for means would be a little risky especially for larger scenes

gsplat/cuda/csrc/world_to_cam_bwd.cu
```diff
@@ -27,8 +27,12 @@ world_to_cam_bwd_kernel(const uint32_t C, const uint32_t N,
                         T *__restrict__ v_covars,         // [N, 3, 3]
                         T *__restrict__ v_viewmats        // [C, 4, 4]
 ) {
+
+    // For now we'll upcast float16 and bfloat16 to float32
+    using OpT = typename OpType<T>::type;
```
liruilong940607, line 32: This is very convenient! It basically maps the data type `T` to `OpT` that defines the precision of the calculation in this kernel.

gsplat/cuda/csrc/types.cuh
```diff
@@ -21,4 +24,24 @@ template <typename T> using mat4 = glm::mat<4, 4, T>;
 
 template <typename T> using mat3x2 = glm::mat<3, 2, T>;
 
+
+template <typename T> struct OpType {
+    typedef T type;
+};
+
+template<> struct OpType<__nv_bfloat16> {
```
liruilong940607, line 32: We also need `OpType<float>` and `OpType<double>` for completeness right?

gsplat/cuda/csrc/world_to_cam_bwd.cu
```diff
@@ -83,7 +90,7 @@ world_to_cam_bwd_kernel(const uint32_t C, const uint32_t N,
             for (uint32_t i = 0; i < 3; i++) { // rows
                 PRAGMA_UNROLL
                 for (uint32_t j = 0; j < 3; j++) { // cols
-                    gpuAtomicAdd(v_covars + i * 3 + j, v_covar[j][i]);
+                    gpuAtomicAdd(v_covars + i * 3 + j, T(v_covar[j][i]));
                 }
```
liruilong940607, line 94: How should we make sure we are not missing all the necessary casting? What kind of tests do we need to add?

gsplat/cuda/csrc/types.cuh
fwilliams: Nope! They just use the default `OpType` definition which is the identity map for these types. i.e.

```c++
template <typename T>
struct OpType {
    typedef T type;
};
```

gsplat/cuda/csrc/world_to_cam_bwd.cu
fwilliams: I'm not 100% sure the explicit cast in all cases. The compiler complains in the atomicAdd case  since the type of `v_covars` is `T*` and `v_covar[j][i]` is `OpT` and there's no matching function definition for `gpuAtomicAdd`. So for all atomic ops, the compiler will catch these for us. 

Otherwise, remark that in general all the outputs are global writes of the form 
```c++
template <typename T>
__global__ kernel(..., T* kernel_output) {
    OpT computed_value = ...;
 
    ...
    
    kernel_output_var[threadIdx] = T(computed_value); 
}
```  

If you don't have the explicit cast on the last line, then the cast is implicitly done, which is what we want anyway. You may get a warning though which you can fix and make the code more readable.

fwilliams:
> cool! will this have support for heterogenous types? eg represent means in float32 and colors/scales in float16? It seems like float16 for means would be a little risky especially for larger scenes

This PR assumes that all floating point inputs to any kernel share the same type. For reference, float16 can represent numbers in the roughly +/- 2^5 with 5 extra bits after the decimal (i.e. +/- 32000.XX) and more bits after the decimal for smaller numbers, which I suspect may be good enough for many scenes. BFloat16 trades more exponent bits for fewer fraction bits.

I think that allowing for mixed precision means is a good idea but would require an extra template parameter and dispatch at the call site of each kernel (you're now switching over 2 types, and compiling 16 kernels, rather than 1 type and 4 kernels). I propose we try the current approach and I can do a second PR for mixed precision?

liruilong940607 (approved):
LGTM!

kerrj:
How would I use this PR? when I pass in half precision inputs to `rasterization` I get `RuntimeError: expected scalar type Float but found Half` from somewhere inside `radii, means2d, depths, conics, compensations = _make_lazy_cuda_func(
            "fully_fused_projection_fwd"
        )`

liruilong940607:
@kerrj This PR only updates half of the functions, which does not include `fully_fused_projection_fwd` I think. @fwilliams is currently doing another PR to update the rest half.

tnarek:
hi! it seems bfloat16 is still not supported? I'm using version 1.5.3 and still getting the same errors when using float16/bfloat16. Any plans to add support?

PR #273, opened 2024-07-08
Implement normal consistency loss.

jefequien:
Implement normal consistency loss for 3DGS-MCMC. It should be a strong baseline for surface reconstruction for only a small drop in PSNR. Hyperparameters are from [2DGS](https://github.com/hbb1/2d-gaussian-splatting) and their implementation is wrapped for comparison.

3DGS

https://github.com/nerfstudio-project/gsplat/assets/6775703/62a867d0-722d-48e2-8642-3e50daebcf49

2DGS
TODO

3DGS-MCMC (3M GSs) wo normal consistency (PSNR: 27.67)
val_step6999_0001

3DGS-MCMC (3M GSs) w normal consistency (PSNR: 27.25)
val_step29999_0001


jefequien:
@liruilong940607 I can clean this PR up and support vanilla 3dgs and packed mode if you guys are interested. Also, I spent some time with distortion loss and didn't see an improvement.

gsplat/cuda/csrc/fully_fused_projection_fwd.cu
```diff
@@ -75,6 +77,9 @@ fully_fused_projection_fwd_kernel(const uint32_t C, const uint32_t N,
         quats += gid * 4;
         scales += gid * 3;
         quat_scale_to_covar_preci<T>(glm::make_vec4(quats), glm::make_vec3(scales), &covar, nullptr);
+
+        glm::mat3 rotmat = quat_to_rotmat(glm::make_vec4(quats));
+        normal = rotmat[2];
```
liruilong940607, line 82: Why is the normal defined like this? This is a bit weird to me. Any paper is defined in this way?

liruilong940607:
Awesome work @jefequien! Certainly this is something very useful! 

I'll come back later for a closer look. BTW any paper are you referencing for this PR? I thought for 3DGS, you would need to implemented the normal formulation either following this paper (https://baowenz.github.io/radegs/) or this one (https://arxiv.org/abs/2404.10772). It doesn't seem like your implementation is one of these but still works like charm. Very interesting.

gsplat/cuda/csrc/fully_fused_projection_fwd.cu
jefequien: I was loosely following the 2DGS implementation and how they read out the last column/row? on the rotation matrix.
Multiplying by the viewmatrix is not necessary because they multiply it on the python side and I felt omitting scale was natural to always read out a unit vector.

```
glm::mat3 R = quat_to_rotmat(rot);
glm::mat3 S = scale_to_mat(scale, mod);
glm::mat3 L = R * S;
normal = transformVec4x3({L[2].x, L[2].y, L[2].z}, viewmatrix);
```
https://github.com/hbb1/diff-surfel-rasterization/blob/28c928a36ea19407cd9754d068bd9a9535216979/cuda_rasterizer/forward.cu#L113

jefequien:
I was actually trying to implement 2DGS-MCMC and this was something much simpler that worked. I haven't had a chance to look at RaDe-GS.

The 2DGS paper says they did something similar in Appendix C under Additional Baselines. Maybe @hbb1 can shed some light here?

> Because the 3D Gaussian lacks a surface normal, we treat its
normal as a trainable parameter [Gao et al. 2023; Liang et al. 2023]
distilled from the depth map, employing the normal consistency
regularization.

liruilong940607:
I think the normal in 2DGS is well defined as it is just a plane. For 3DGS the definition of normal is very ambiguous. The normal for 3DGS probably have to originate from the ray-GS intersection plane, like whats been done in the above two papers.

jefequien:
Yeah. It's pretty surprising. It's just read out and then accumulated like a color.

hbb1:
@jefequien I give a learnable normal vector to each 3D Gaussian and employ the same Normal-depth regularization. This is similar as those works for 3DGS' inverse rendering, except that I don't detach the weight of the depth-map. It will both smooth the depth map and learn accurate normal for visualization.

jefequien:
Oh. So you used an additional three learnable parameters for each Gaussian. I'm not sure I follow "those works for 3DGS' inverse rendering". Do you have any links?

I also didn't ablate the detach on `normal * (alpha).detach()`. That's copied from 2DGS. 
https://github.com/hbb1/2d-gaussian-splatting/blob/61c7b417393d5e0c58b742ad5e2e5f9e9f240cc6/gaussian_renderer/__init__.py#L147

hbb1:
You can check  papers like [GShader](https://github.com/Asparagus15/GaussianShader/blob/main/gaussian_renderer/__init__.py) and [Relit3DGS](https://github.com/NJU-3DV/Relightable3DGaussian)
So did you directly use the last row of the rotation matrix as normal? 😂  I wonder if  it makes 3D Gaussian flatten along the third axis. Then that would make sense.

examples/simple_trainer_mcmc.py
```diff
@@ -433,15 +459,22 @@ def train(self):
...
+                depths = renders[..., -1:]
+                normals = renders[..., -4:-1]
+                normals_surf = depth_to_normal(
+                    depths,
+                    camtoworlds,
+                    Ks,
+                    near_plane=cfg.near_plane,
+                    far_plane=cfg.far_plane,
+                )
+                normals_surf = normals_surf * (alphas).detach()
```
niujinshuchong, line 472: Why do you need to multiply with alphas? Isn’t normals_surf already normalized here: https://github.com/jefequien/gsplat-mcmc/blob/6d156d2d447d9557e432b7f4b94fa6b3b9527187/gsplat/normal_utils.py#L41?

gsplat/normal_utils.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def _depth_to_normal(depth, world_view_transform, full_proj_transform):
+    points = _depths_to_points(
+        depth, world_view_transform, full_proj_transform
+    ).reshape(*depth.shape[:2], 3)
+    output = torch.zeros_like(points)
+    dx = torch.cat([points[2:, 1:-1] - points[:-2, 1:-1]], dim=0)
+    dy = torch.cat([points[1:-1, 2:] - points[1:-1, :-2]], dim=1)
+    normal_map = torch.nn.functional.normalize(torch.cross(dx, dy, dim=-1), dim=-1)
```
niujinshuchong, line 41: Normal from depth map is normalized.

gsplat/cuda/csrc/fully_fused_projection_fwd.cu
niujinshuchong: Why the normal don't need to accumulated in rasterize_to_pixels_fwd.cu?

jefequien:
This PR is also similar to 2.5d-gaussian-splatting in using the row of the rotation matrix as the normal vector but I don't explicitly set the Gaussian z scale to zero.
https://github.com/hugoycj/2.5d-gaussian-splatting/blob/e2c8ad0858be3a6d4ea266d00f0f228e9d6e3a03/gaussian_renderer/__init__.py#L121

examples/simple_trainer_mcmc.py
jefequien: You can either normalize both the rendered normals and surface normals or use the alpha-composited rendered normals and multiply the surface normals by alpha. Doing the latter looks better visually.

gsplat/cuda/csrc/fully_fused_projection_fwd.cu
jefequien: They are. Normals are concatenated onto color and accumulated with them. In this PR, COLOR_DIM is 7. 3 for color, 3 for normal, and 1 for depth.
https://github.com/jefequien/gsplat-mcmc/blob/jeff/normal_consistency/gsplat/rendering.py#L320

gsplat/normal_utils.py
jefequien: The normal map from a depth map should always be normalized right?
https://github.com/hbb1/2d-gaussian-splatting/blob/main/utils/point_utils.py#L35

examples/simple_trainer_mcmc.py
niujinshuchong: Thanks for the clarification. I was using the first one in GOF.

gsplat/normal_utils.py
hbb1: The latter one should be more robust when dealing semi-transparent pixels.

examples/simple_trainer_mcmc.py
hbb1: The latter one should be more robust when dealing with semi-transparent pixels (accumulate_alpha < 1).

LearnedVector:
Hi, are there plans to merge this into the main branch?

hiyyg:
any updates?

PR #274, opened 2024-07-09
[DRAFT] Mip-Splatting implementation

niujinshuchong:
Implementation of [Mip-Splatting](https://github.com/autonomousvision/mip-splatting). Tested on bicycle scenes and works OK.

Here is a comparison when trained with factor 4 and render with factor 1.

Needs more testing and clean up later.

### DIFF
--- a/examples/simple_trainer_mip_splatting.py
+++ b/examples/simple_trainer_mip_splatting.py
@@ -1,0 +1,1069 @@
+import json
+import math
+import os
+import time
+from dataclasses import dataclass, field
+from typing import Dict, List, Optional, Tuple
+
+import imageio
+import numpy as np
+import torch
+import torch.nn.functional as F
+import tqdm
+import tyro
+import viser
+import nerfview
+from datasets.colmap import Dataset, Parser
+from datasets.traj import generate_interpolated_path
+from torch import Tensor
+from torch.utils.tensorboard import SummaryWriter
+from torchmetrics.image import PeakSignalNoiseRatio, StructuralSimilarityIndexMeasure
+from torchmetrics.image.lpip import LearnedPerceptualImagePatchSimilarity
+from utils import (
+    AppearanceOptModule,
+    CameraOptModule,
+    knn,
+    normalized_quat_to_rotmat,
+    rgb_to_sh,
+    set_random_seed,
+)
+
+from gsplat.rendering import rasterization
+
+
+@dataclass
+class Config:
+    # Disable viewer
+    disable_viewer: bool = False
+    # Path to the .pt file. If provide, it will skip training and render a video
+    ckpt: Optional[str] = None
+
+    # Path to the Mip-NeRF 360 dataset
+    data_dir: str = "data/360_v2/garden"
+    # Downsample factor for the dataset
+    data_factor: int = 4
+    # Directory to save results
+    result_dir: str = "results/garden"
+    # Every N images there is a test image
+    test_every: int = 8
+    # Random crop size for training  (experimental)
+    patch_size: Optional[int] = None
+    # A global scaler that applies to the scene size related parameters
+    global_scale: float = 1.0
+
+    # Port for the viewer server
+    port: int = 8080
+
+    # Batch size for training. Learning rates are scaled automatically
+    batch_size: int = 1
+    # A global factor to scale the number of training steps
+    steps_scaler: float = 1.0
+
+    # Number of training steps
+    max_steps: int = 30_000
+    # Steps to evaluate the model
+    eval_steps: List[int] = field(default_factory=lambda: [7_000, 30_000])
+    # Steps to save the model
+    save_steps: List[int] = field(default_factory=lambda: [7_000, 30_000])
+
+    # Initialization strategy
+    init_type: str = "sfm"
+    # Initial number of GSs. Ignored if using sfm
+    init_num_pts: int = 100_000
+    # Initial extent of GSs as a multiple of the camera extent. Ignored if using sfm
+    init_extent: float = 3.0
+    # Degree of spherical harmonics
+    sh_degree: int = 3
+    # Turn on another SH degree every this steps
+    sh_degree_interval: int = 1000
+    # Initial opacity of GS
+    init_opa: float = 0.1
+    # Initial scale of GS
+    init_scale: float = 1.0
+    # Weight for SSIM loss
+    ssim_lambda: float = 0.2
+
+    # Near plane clipping distance
+    near_plane: float = 0.01
+    # Far plane clipping distance
+    far_plane: float = 1e10
+
+    # GSs with opacity below this value will be pruned
+    prune_opa: float = 0.005
+    # GSs with image plane gradient above this value will be split/duplicated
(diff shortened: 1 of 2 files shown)


examples/simple_trainer_mip_splatting.py
```diff
@@ -0,0 +1,1069 @@
...
+        opacities = torch.sigmoid(self.splats["opacities"])  # [N,]
+        filters = self.splats["filters"]  # [N,]
+        
+        # apply 3D smoothing filter to scales and opacities
+        scales_square = torch.square(scales) # [N, 3]
+        det1 = scales_square.prod(dim=1)     # [N, ]
+        
+        scales_after_square = scales_square + torch.square(filters) [:, None] # [N, 1]
+        det2 = scales_after_square.prod(dim=1) # [N,]
+        coef = torch.sqrt(det1 / det2) # [N,]
```
jb-ye, line 371: Could the scale be very small, sqrt is not differentiable near zero. Maybe consider add a clamp to avoid zero differentiation?

liruilong940607:
Awesome @niujinshuchong ! 

I'm thinking if there is a way to elegantly message some of implementation into the gsplat library, e.g., into `rasterization()` function so that the downstream users (e.g., nerfstudio) can just import and use.

It doesn't have to be living in the `rasterization()` function if it does not fit there. We could also put some implementation in to like `gsplat/mip.py`, so that ppl can use it with:
```
from gsplat.mip import ...
# and do their thing
```

liruilong940607:
I'll come back to it later with a closer look. Great job!

niujinshuchong:
@liruilong940607 Thanks for the suggestions. I added a cuda implementation of compute_3D_smoothing_filter and it can be imported from gsplat. The benchmark results (see top) are reasonable. For now, I create a new script `simple_trainer_mip_splatting.py` and it almost the same as `simple_trainer.py`, do you think we should merge them or make it separate? I also added a training script to use multiple gpus.

examples/simple_trainer_mip_splatting.py
```diff
@@ -0,0 +1,1094 @@
...
+        height, width = data["image"].shape[:2]
+        Ks = torch.stack([K] * len(worldtocams), dim=0)  # [C, 3, 3]
+        filter_3D = compute_3D_smoothing_filter(
+            xyz, worldtocams, Ks, width, height, cfg.near_plane
+        )
+        filter_3D = filter_3D * (0.2**0.5)
+        self.splats["filters"] = torch.nn.Parameter(filter_3D)
+
+    @torch.no_grad()
+    def compute_3D_smoothing_filter_torch(self):
```
liruilong940607, line 709: Would you mind move this into `gsplat/cuda/_torch_impl.py` and put up a test case under `test/test_basic.py` to protect the CUDA implementation `compute_3D_smoothing_filter()`?

liruilong940607:
Thanks for the benchmarking! Really nice! Yeah from a quick going through of the training script, it seems to be very compatible with the original one so I would suggest we merge the two! Just need a flag in the config, e.g.(--mip_splatting?) and a couple if else in the code.

gsplat/cuda/_wrapper.py
```diff
@@ -572,6 +572,34 @@ def rasterize_to_indices_in_range(
     return out_gauss_ids, out_pixel_ids, out_camera_ids
 
 
+def compute_3D_smoothing_filter(
```
liruilong940607, line 575: nit: maybe use `3d` instead of `3D` in the naming of these functions to align with the naming convention in gsplat?

examples/benchmark_mipnerf360.py
liruilong940607, line 1: Nice scripts! 
- Does `GPUtil` need to be installed via pip? If so we should put it into `examples/requirements.txt`.
- If you are going to merge the `simple_trainer_mip_splatting.py` with `simple_trainer.py` then we can just replace the original benchmark script with this one, and enable an input flag to enable mip splatting. But merging the benchmark script means that we probably want the default benchmarking set to be the original 3DGS.

examples/benchmark_mipnerf360_stmt.py
```diff
@@ -0,0 +1,115 @@
...
+# factors = [4, 2, 2, 4, 4, 4, 4, 2, 2]
+
+# 7 scenes
+scenes = ["bicycle", "bonsai", "counter", "garden", "stump", "kitchen", "room"]
+factors = [8] * len(scenes)
+
+excluded_gpus = set([])
+
+# classic
+result_dir = "results/benchmark_stmt"
```
liruilong940607, line 21: what does `stmt` stands for? lol

examples/benchmark_mipnerf360_stmt.py
liruilong940607, line 1: how hard would it be to merge this one with the other script? Ideally an list argument could be passed in to specify the rendering factors that the evaluation happens on

examples/show_mipnerf360.py
liruilong940607, line 1: Nice script! Would love to have it! Maybe the name of this script could be something like "compare_results_mipnerf360" to be more readable? Actually maybe we could also get rid of the `_mipnerf360` suffix until when another dataset is supported.

examples/show_mipnerf360_allscales.py
```diff
@@ -0,0 +1,45 @@
+import json
```
liruilong940607, line 1: actually, another idea of managing these scripts is to have a `mip_splatting` folder under `examples`, in which you could have multiple scripts in there, without worrying about merging or conflicting with other scripts. I think this is a scalable way to manage the scripts when more things are supported in the future.

And you could also have a customized `simple_trainer.py` living in it, basically whatever you need to be customized. But this is less preferred because a customized `simple_trainer.py` means the mip-splatting feature would be disjoint with other future features for user to play with. So ideally we merge the trainer but could put the scripts into an isolated folder.

examples/benchmark_mipnerf360_stmt.py
maturk: I think` Single-scale-Training-and-Multi-scale-Testing`, the idea being you train with a fixed resolution (here a factor of 1/8 of the original image res) and test on various other resolutions (1x full res,  1/2, 1/4, 1/8).

examples/benchmark_mipnerf360_stmt.py
maturk: I think we could adapt the current `benchmark.sh` script into the `.py` style similar to the one proposed here, allowing for easier batch jobs on e.g. clusters and also streamlining the evaluation of the various new features of gsplat... I can look into it...

jb-ye:
What's the plan on merging this PR? I am particularly interested in this feature (esp. the 3D filter part).

liruilong940607:
I think this PR is just about cleaning up, add doc strings and update webpage. It would be nice to finish it up.

Not sure if @niujinshuchong still has time for it. But I might be able to find some cycles in the next two weeks to help with finishing up this PR.

niujinshuchong:
@liruilong940607 Yes, it just need to be clean up. I will find some time to continue. However, it would be great if you can help with it. Sorry for the late reply.

PR #278, opened 2024-07-11
Abstracting out the GS heuristics into a Strategy class

liruilong940607:
The idea is to move the densification heuristics from the example scripts into gsplat library and package them into a class, so that the downstream users could simply use them via `from gsplat.strategy import DefaultStrategy, MCMCStrategy`.  

The API is fairly simple:
```
params: torch.nn.ParameterDict = ...
optimizers: Dict[str, torch.optim.Optimizer] = ...
strategy = DefaultStrategy(<arguments>)
state = strategy.initialize_state()

for step in range(max_steps):
    render_color, render_alpha, info = rasterization(...)
    strategy.step_pre_backward(params, optimizers, state, step=step, info=info, <other arguments>)
    loss.backward()
    strategy.step_post_backward(params, optimizers, state, step=step, info=info, <other arguments>)
```

This would also allow users to switch between different strategies, and would make it more easily to embrace more GS heuristics. This is a joint efforts with @rtabrizi.

I also verified that by switch to `DefaultStrategy` and `MCMCStrategy` in the example script, nothing is affected. Now these two scripts have very little differences so we could potentially merge them into one script. But I'll leave it to another PR to minimize the code change of this PR

```
python simple_trainer.py

# Before this PR
> Step:  6999 {'mem': 7.582746505737305, 'ellipse_time': 424.46320390701294, 'num_GS': 4488772}
> {"psnr": 26.358938217163086, "ssim": 0.8353559374809265, "lpips": 0.12293213605880737, "ellipse_time": 0.026054859161376953, "num_GS": 4488772}

# After this PR
> Step:  6999 {'mem': 7.579144477844238, 'ellipse_time': 423.3516597747803, 'num_GS': 4481712}
> {"psnr": 26.348445892333984, "ssim": 0.8354260325431824, "lpips": 0.12293919920921326, "ellipse_time": 0.026077985763549805, "num_GS": 4481712}
```

```
python simple_trainer_mcmc.py 

# Before this PR
> Step:  6999 {'mem': 1.7906079292297363, 'ellipse_time': 247.97816705703735, 'num_GS': 1000000}
> {"psnr": 26.010623931884766, "ssim": 0.8161215782165527, "lpips": 0.15706905722618103, "ellipse_time": 0.009712855021158854, "num_GS": 1000000}

# After this PR
> Step:  6999 {'mem': 1.7903776168823242, 'ellipse_time': 239.95758247375488, 'num_GS': 1000000}
> {"psnr": 26.001811981201172, "ssim": 0.8158013224601746, "lpips": 0.1573895812034607, "ellipse_time": 0.009663820266723633, "num_GS": 1000000}
```

TODO:

- [x] Documentation. @liruilong940607 
- [x] See if it fits with Splatfacto in nerfstudio. [branch in nerfstudio](https://github.com/nerfstudio-project/nerfstudio/tree/splatfacto-refactor)

tentative doc preview: https://669107b6a724ef1f2c186e67--inquisitive-speculoos-27b42b.netlify.app/apis/strategy


gsplat/strategy.py
```diff
@@ -0,0 +1,552 @@
...
+from gsplat.utils import normalized_quat_to_rotmat
+
+
+class Strategy(ABC):
+    """Base class for the GS optimization strategy."""
+
+    def __init__(
+        self,
+        params: torch.nn.ParameterDict,
+        optimizers: List[torch.optim.Optimizer],
```
kerrj, line 20: wouldn't this need to be a dict matching the key in the paramdict to the optimizer?

gsplat/strategy.py
```diff
@@ -0,0 +1,552 @@
...
+        # Start refining GSs after this iteration
+        refine_start_iter: int = 500,
+        # Stop refining GSs after this iteration
+        refine_stop_iter: int = 15_000,
+        # Reset opacities every this steps
+        reset_every: int = 3000,
+        # Refine GSs every this steps
+        refine_every: int = 100,
+        # Use absolute gradients for GS splitting
+        absgrad: bool = False,
```
kerrj, line 84: if this is true maybe we want to automatically set the grow_grad2d to .008 otherwise the number of gaussians will explode? or maybe warn the user?

gsplat/strategy.py
liruilong940607: A single optimizer is enough as long as the param group has a "name" field defined. The reason I'm putting a List here is in case user wants to use different optimizer for different field. E.g., SGD for color and Adam for the rest

https://github.com/nerfstudio-project/gsplat/blob/main/examples/simple_trainer.py#L217-L224

gsplat/strategy.py
```diff
@@ -0,0 +1,552 @@
...
+            self._state["grad2d"].index_add_(0, gs_ids, grads[sel].norm(dim=-1))
+            self._state["count"].index_add_(
+                0, gs_ids, torch.ones_like(gs_ids, dtype=torch.float32)
+            )
+
+    @torch.no_grad()
+    def _refine_duplicate(self, mask: Tensor):
+        device = mask.device
+        sel = torch.where(mask)[0]
+        for optimizer in self._optimizers:
```
kerrj, line 248: do we want some sort of helper function for this? looping through optimizers and adding/pruning/removing is a common pattern that all strategies will need

kerrj: so these lines could then call `self.duplicate_in_optimizer(index_mask)` or something

gsplat/strategy.py
liruilong940607: Not sure if warning is a good idea. on different scenes this threshold could be very different. Auto set is also not ideal as we want to give the user the freedom to adjust that. 

I think the best we can do is probably note it clearly in the docstring.

gsplat/strategy.py
kerrj: I see, so it will automatically loop through all optimizers and check if any of the param group names match the paramdict?

gsplat/strategy.py
liruilong940607: I tried once but didn't end up creating anything elegant. Open for suggestions!

gsplat/strategy.py
kerrj: in nerfstudio there is a separate `torch.nn.Optimizer` for each param group, so we should check if passing in a list of those works

gsplat/strategy.py
kerrj: [These functions](https://github.com/nerfstudio-project/nerfstudio/blob/27801d2a4786de0f94c280c0fbf27ac717dd0d22/nerfstudio/models/splatfacto.py#L373) in splatfacto try to do something like this, so any subclass could also call dup_in_optim and have the same effect

kerrj: I think the body of the function would need to be different but the general interface of taking in a per-gaussian mask could work for removing/duplicating, or specifying how many to append

gsplat/strategy.py
liruilong940607: List of optimizers also works as long as the param group has a "name" field.

The logic is that it will loop over the list of optimizers and every parameter group in it to find a match, based on the "name" field

gsplat/strategy.py
liruilong940607: The nerfstudio way of implementation is mixing the duplication, splitting, and removing operations together so that the remove on optimizer can be done once. But the downside is that it's hard to read and modify. I would like to keep each operation isolated. But yeah maybe there is a way to share some code between them. I just dont have a concrete idea.

gsplat/strategy.py
liruilong940607: Or, another idea is to create another class `AbsGradStrategy` that inherit from this class and set the default value to 0.008.

gsplat/strategy.py
liruilong940607: Actually, maybe its more user friendly to just use a dict optimizer and force each parameter has a seperate optimizer?

brentyi (commented):
I left some very minor comments!

One overall suggestion that I'm personally drawn to (but am not sure that others would be) and consolidates some of my review comments is to consider a stateless `Strategy` object:
```python
class Strategy(abc.ABC):
    @abc.abstractmethod
    def initialize_state(self) -> Dict[str, Any]:
        ...  # Return some initial state. Doesn't do any `self.xxx = ...` assignment.

    @abc.abstractmethod
    def step_pre_backward(
        self,
        state: Dict[str, Any],  # Take the state explicitly as input.
        parameters: torch.nn.ParameterDict,
        optimizers: List[torch.nn.Optimizers],
        step: int,
        info: Dict[str, Any],
    ) -> None:
        ...

    @abc.abstractmethod
    def step_post_backward(...) -> None:  # Same inputs as `step_pre_backward()`!
        ...
```

It'd be more verbose where the callbacks are called but:
- The main thing I'm a fan of is the strategy object defining behavior without itself needing to "own" any optimizers/tensors/etc. The cognitive overhead of using it would just be lower to me; the strategy objects are otherwise tied to a specific set of params + torch devices
- It would make some aspects of subclassing easier, like less boilerplate associated with the superclass arguments
- Instantiation would be more flexible for users. Currently the strategy can't be instantiated until after all of the parameters and optimizers are set up.
    - If you wanted to compare the strategies and write some loop like `for strategy in [DefaultStrategy(...), MCMCStrategy(...)]: train_gs_and_print_stats(strategy)` it would be easier
    - We could also simplify the simple trainer by replacing all of the repeated strategy arguments in the trainer config with just a nested object `strategy: DefaultStrategy` (especially if `DefaultStrategy` is a dataclass)

liruilong940607:
> I left some very minor comments!
> 
> One overall suggestion that I'm personally drawn to (but am not sure that others would be) and consolidates some of my review comments is to consider a stateless `Strategy` object:
> 
> ```python
> class Strategy(abc.ABC):
>     @abc.abstractmethod
>     def initialize_state(self) -> Dict[str, Any]:
>         ...  # Return some initial state. Doesn't do any `self.xxx = ...` assignment.
> 
>     @abc.abstractmethod
>     def step_pre_backward(
>         self,
>         state: Dict[str, Any],  # Take the state explicitly as input.
>         parameters: torch.nn.ParameterDict,
>         optimizers: List[torch.nn.Optimizers],
>         step: int,
>         info: Dict[str, Any],
>     ) -> None:
>         ...
> 
>     @abc.abstractmethod
>     def step_post_backward(...) -> None:  # Same inputs as `step_pre_backward()`!
>         ...
> ```
> 
> It'd be more verbose where the callbacks are called but:
> 
> * The main thing I'm a fan of is the strategy object defining behavior without itself needing to "own" any optimizers/tensors/etc. The cognitive overhead of using it would just be lower to me; the strategy objects are otherwise tied to a specific set of params + torch devices
> * It would make some aspects of subclassing easier, like less boilerplate associated with the superclass arguments
> * Instantiation would be more flexible for users. Currently the strategy can't be instantiated until after all of the parameters and optimizers are set up.
>   
>   * If you wanted to compare the strategies and write some loop like `for strategy in [DefaultStrategy(...), MCMCStrategy(...)]: train_gs_and_print_stats(strategy)` it would be easier
>   * We could also simplify the simple trainer by replacing all of the repeated strategy arguments in the trainer config with just a nested object `strategy: DefaultStrategy` (especially if `DefaultStrategy` is a dataclass)

Ok I like this a lot! Also makes it painless to integrate into nerfstudio. I'll update to this way

gsplat/strategy/default.py
```diff
@@ -0,0 +1,334 @@
...
+        self._prune_opa = prune_opa
+        self._grow_grad2d = grow_grad2d
+        self._grow_scale3d = grow_scale3d
+        self._prune_scale3d = prune_scale3d
+        self._refine_start_iter = refine_start_iter
+        self._refine_stop_iter = refine_stop_iter
+        self._reset_every = reset_every
+        self._refine_every = refine_every
+        self._absgrad = absgrad
+        self._revised_opacity = revised_opacity
```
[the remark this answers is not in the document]
liruilong940607: Oh nice! I'm just not that familiar with dataclass so it didn't ring a bell to me. I'll update this so that the nested config could work. Just curious, are there another advantages that dataclass would bring to us (in this case)?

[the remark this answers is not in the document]
brentyi: I think the convenient constructor, subclassing convenience, being able to use it as a config object are the main things!

There are other benefits of dataclasses (introspection, hashing, comparison, easy immutability via `frozen=`, etc) but I don't think they really matter here.

gsplat/relocation.py
```diff
@@ -1,20 +1,19 @@
...
+from .cuda._wrapper import _make_lazy_cuda_func
 
 
 def compute_relocation(
     opacities: Tensor,  # [N]
     scales: Tensor,  # [N, 3]
     ratios: Tensor,  # [N]
-) -> tuple[Tensor, Tensor]:
+    binoms: Tensor,  # [n_max, n_max]
+    n_max: int,
```
liruilong940607, line 15: @jefequien Would you mind helping me write the doc string for these two? I'm not sure how to best describe them. Dumping your comments here is fine I can add it to the code!

liruilong940607:
Preview Doc is online: https://669107b6a724ef1f2c186e67--inquisitive-speculoos-27b42b.netlify.app/apis/strategy

gsplat/relocation.py
jefequien: binoms: precomputed lookup table for binomial coefficients used in Equation 9 in "3D Gaussian Splatting as Markov Chain Monte Carlo".
n_max: size of lookup table for binomial coefficients. Clamp values larger than this to prevent out-of-bounds exceptions.
```
{'binoms': tensor([[   1,    0,    0,  ...,    0,    0,    0],
        [   1,    1,    0,  ...,    0,    0,    0],
        [   1,    2,    1,  ...,    0,    0,    0],
        ...,
        [   1,   48, 1128,  ...,    1,    0,    0],
        [   1,   49, 1176,  ...,   49,    1,    0],
        [   1,   50, 1225,  ..., 1225,   50,    1]]), 'n_max': 51}
```
You can probably remove n_max and use binoms.size(0) on the cuda side.

docs/source/apis/strategy.rst
```diff
@@ -0,0 +1,76 @@
+Densification Strategy
+===================================
+
+.. currentmodule:: gsplat
+
+In `gsplat`, we abstract out the densification and pruning process of the Gaussian 
+training into a strategy. A strategy is a class that defines how the Gaussian parameters
+(along with their optimizers) should be updated (splitting, pruning etc) during the training.
```
rtabrizi, line 8: nit: (splitting, pruning etc) --> (splitting, pruning, etc.)

docs/source/apis/strategy.rst
```diff
@@ -0,0 +1,76 @@
...
+        strategy.step_post_backward(params, optimizers, strategy_state, step, info)
+
+A strategy will inplacely update the Gaussian parameters as well as the optimizers,
+so it has a specific expectation on the format of the parameters and the optimizers.
+It is designed to work with the Guassians defined as either a Dict of
+`torch.nn.Parameter <https://pytorch.org/docs/stable/generated/torch.nn.parameter.Parameter.html>`_
+or a
+`torch.nn.ParameterDict <https://pytorch.org/docs/stable/generated/torch.nn.ParameterDict.html>`_
+with at least the following keys: {"means", "scales", "quats", "opacities"}. On top of these attributes,
+arbitrarily number of extra attributes are supported. Besides the parameters, it also
```
rtabrizi, line 53:
```suggestion
an arbitrary number of extra attributes are supported. Besides the parameters, it also
```

PR #284, opened 2024-07-14
Add optional masks in `rasterize_to_pixels()` to support Grendel

alexis-mmm:
Now `rasterize_to_pixels()` supports an optional argument `masks` to identify its local workload in a multi-GPU scenario, so it can better work with [Grendel](https://github.com/nyu-systems/Grendel-GS). When a tile is marked as false:
- In the forward pass, all channels of each pixel in that tile are set to zero.
- In the backward pass, no operations are performed on that tile.

### DIFF
--- a/gsplat/cuda/_wrapper.py
+++ b/gsplat/cuda/_wrapper.py
@@ -379,6 +379,7 @@ def rasterize_to_pixels(
     isect_offsets: Tensor,  # [C, tile_height, tile_width]
     flatten_ids: Tensor,  # [n_isects]
     backgrounds: Optional[Tensor] = None,  # [C, channels]
+    masks: Optional[Tensor] = None,  # [C, tile_height, tile_width]
     packed: bool = False,
     absgrad: bool = False,
 ) -> Tuple[Tensor, Tensor]:
@@ -395,6 +396,7 @@ def rasterize_to_pixels(
         isect_offsets: Intersection offsets outputs from `isect_offset_encode()`. [C, tile_height, tile_width]
         flatten_ids: The global flatten indices in [C * N] or [nnz] from  `isect_tiles()`. [n_isects]
         backgrounds: Background colors. [C, channels]. Default: None.
+        masks: Optional masks to support Grendel's local workload strategy. [C, tile_height, tile_width]. Default: None.
         packed: If True, the input tensors are expected to be packed with shape [nnz, ...]. Default: False.
         absgrad: If True, the backward pass will compute a `.absgrad` attribute for `means2d`. Default: False.
 
@@ -422,6 +424,9 @@ def rasterize_to_pixels(
     if backgrounds is not None:
         assert backgrounds.shape == (C, colors.shape[-1]), backgrounds.shape
         backgrounds = backgrounds.contiguous()
+    if masks is not None:
+        assert masks.shape[0] == C, masks.shape
+        masks = masks.contiguous()
 
     # Pad the channels to the nearest supported number if necessary
     channels = colors.shape[-1]
@@ -484,6 +489,7 @@ def rasterize_to_pixels(
         colors.contiguous(),
         opacities.contiguous(),
         backgrounds,
+        masks,
         image_width,
         image_height,
         tile_size,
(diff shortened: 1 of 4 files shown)


gsplat/cuda/csrc/rasterize_to_pixels_bwd.cu
```diff
@@ -52,6 +53,15 @@ __global__ void rasterize_to_pixels_bwd_kernel(
     if (backgrounds != nullptr) {
         backgrounds += camera_id * COLOR_DIM;
     }
+    if (masks != nullptr) {
+        masks += camera_id * tile_height * tile_width;
+    }
+
+    // when the mask is provided, do nothing and return if
+    // this tile is labeled as not local
```
liruilong940607, line 61: nit: "local" is a bit unclear what it means if ppl do not have a background knowledge about grendel. Maybe just say "if this tile is labeled as False"

gsplat/cuda/csrc/rasterize_to_pixels_fwd.cu
```diff
@@ -53,6 +57,15 @@ __global__ void rasterize_to_pixels_fwd_kernel(
     bool inside = (i < image_height && j < image_width);
     bool done = !inside;
 
+    // when the mask is provided, render 0.0 and return if this tile
```
liruilong940607, line 60: I think this conflicts with when `backgrounds` is provided. In the case of a background is provided, user might expect the background color to be rendered at masked region.

gsplat/cuda/csrc/rasterize_to_pixels_fwd.cu
```diff
@@ -53,6 +57,15 @@ __global__ void rasterize_to_pixels_fwd_kernel(
...
+    // is labeled as not local
+    if (masks != nullptr && inside && !masks[tile_id]) {
+        for (uint32_t k = 0; k < COLOR_DIM; ++k) {
+            render_colors[pix_id * COLOR_DIM + k] = 0.0f;
```
liruilong940607, line 64: If background is enabled, should set color as background

gsplat/cuda/_wrapper.py
```diff
@@ -395,6 +396,7 @@ def rasterize_to_pixels(
         isect_offsets: Intersection offsets outputs from `isect_offset_encode()`. [C, tile_height, tile_width]
         flatten_ids: The global flatten indices in [C * N] or [nnz] from  `isect_tiles()`. [n_isects]
         backgrounds: Background colors. [C, channels]. Default: None.
+        masks: Optional masks to support Grendel's local workload strategy. [C, tile_height, tile_width]. Default: None.
```
liruilong940607, line 399: Same here, would be quite confused to user if the user is not familiar with grendel. Should just say "Optional tile mask to skip render GS to some tiles."

gsplat/cuda/_wrapper.py
```diff
@@ -422,6 +424,9 @@ def rasterize_to_pixels(
     if backgrounds is not None:
         assert backgrounds.shape == (C, colors.shape[-1]), backgrounds.shape
         backgrounds = backgrounds.contiguous()
+    if masks is not None:
+        assert masks.shape[0] == C, masks.shape
```
liruilong940607, line 428: maybe better to check it to be the same shape with `isect_offsets`

liruilong940607 (commented):
Nice! Added some minor comments

gsplat/cuda/csrc/rasterize_to_pixels_fwd.cu
alexis-mmm: Agree! Though it's not required in Grendel, rendering the background color at masked region is generally useful in many other cases to improve computational efficiency. Will fix this.

gsplat/cuda/csrc/rasterize_to_pixels_bwd.cu
alexis-mmm: fixed

gsplat/cuda/csrc/rasterize_to_pixels_fwd.cu
alexis-mmm: fixed

gsplat/cuda/csrc/rasterize_to_pixels_fwd.cu
alexis-mmm: fixed

gsplat/cuda/_wrapper.py
alexis-mmm: fixed

gsplat/cuda/_wrapper.py
alexis-mmm: fixed

liruilong940607 (approved):
LGTM now! Thanks for contributing!

PR #305, opened 2024-07-27
Fix projection for images with non-centered camera (e.g. crops)

JonathonLuiten:
The current code assumes [cx,cy] are at [w/2, h/2], if this is not the case (very common if you want to render a non-centered crop) the code gives VERY bad and incorrect results. This quick fix uses the proper [cx, cy].

image

### DIFF
--- a/gsplat/cuda/_torch_impl.py
+++ b/gsplat/cuda/_torch_impl.py
@@ -83,13 +83,17 @@ def _persp_proj(
 
     fx = Ks[..., 0, 0, None]  # [C, 1]
     fy = Ks[..., 1, 1, None]  # [C, 1]
+    cx = Ks[..., 0, 2, None]  # [C, 1]
+    cy = Ks[..., 1, 2, None]  # [C, 1]
     tan_fovx = 0.5 * width / fx  # [C, 1]
     tan_fovy = 0.5 * height / fy  # [C, 1]
 
-    lim_x = 1.3 * tan_fovx
-    lim_y = 1.3 * tan_fovy
-    tx = tz * torch.clamp(tx / tz, min=-lim_x, max=lim_x)
-    ty = tz * torch.clamp(ty / tz, min=-lim_y, max=lim_y)
+    lim_x_pos = (width - cx) / fx + 0.3 * tan_fovx
+    lim_x_neg = cx / fx + 0.3 * tan_fovx
+    lim_y_pos = (height - cy) / fy + 0.3 * tan_fovy
+    lim_y_neg = cy / fy + 0.3 * tan_fovy
+    tx = tz * torch.clamp(tx / tz, min=-lim_x_neg, max=lim_x_pos)
+    ty = tz * torch.clamp(ty / tz, min=-lim_y_neg, max=lim_y_pos)
 
     O = torch.zeros((C, N), device=means.device, dtype=means.dtype)
     J = torch.stack(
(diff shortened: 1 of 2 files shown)


maturk (approved):
Wow, nice. LGTM!

liruilong940607:
Any reason the backward (persp_proj_vjp) is not updated?

jb-ye:
This seems a major bug. Could you also fix the [backward pass](https://github.com/nerfstudio-project/gsplat/blob/main/gsplat/cuda/csrc/utils.cuh#L191-L192) as well.

hbb1:
can we do so instead, without modifying the rasterizer? https://github.com/graphdeco-inria/gaussian-splatting/issues/144#issuecomment-1938504456

f-dy:
Can someone explain why the clamping?

The original 3DGS code has it too, and a comment says:
```
// The following models the steps outlined by equations 29
// and 31 in "EWA Splatting" (Zwicker et al., 2002).
// Additionally considers aspect / scaling of viewport.
// Transposes used to account for row-/column-major conventions.
```

But EWA splatting has no clamping.

What it does is that it clamps the 2D expansion (due to the projection) of Gaussians that are outside of the view frustum. My guess is that it's something that was introduced at some point, maybe to fix a visual artifact, and left there for no reason (maybe that artifact was due to something else).

I can understand the reason for clamping, when considering very large Gaussians: The projection cannot be considered to be affine anymore, and maybe the clamping was introduced to avoid very large Gaussians that are outside of the view frustum to come inside the view frustum because of the affine approximation, which is computed at the center of the Gaussian and not at the "side" of the Gaussian that is closer to the view frustum. But then why treat Gaussians that are outside and not those that are inside? The projection of Gaussians that are inside should actually be larger (that same Jacobian is higher on the sides of Gaussians that are at the principal point).

I think the correct fix is to remove the clamping.

JonathonLuiten:
@liruilong940607 yes this probably needs to be changed in the backward pass too! I don't have time to look into this myself right now though unfortunately.

@hbb1 The function you linked doesn't fix this issue, this actually needs to be changed in the cuda like suggested in this PR.

@f-dy 'Why clamping': If you do nothing, the 'approximate jacobian projection' of Gaussians out to the side of field of view make these (normally small) Gaussians very large and project into the field of view and results in complete mess (no good image at all), so we absolutely need to do something about this.

There are two solutions:
a) either completely remove Gaussians far out of field of view, e.g. Gaussians outside of 1.3*fov are completely removed / culled. This is a fine solution imo.

b) (done here) make the Gaussians outside of field of view smaller (e.g. clip their size to the size it would be if they were 1.3*fov away). The advantage of this over (a) is that if you have VERY large Gaussians outside of 1.3*fov that actually SHOULD contribute to the image, in this setting they are able to, whereas if you just completely removed them as in (a) they wouldn't be able to contribute.

Note that in (b) there is still culling of Gaussians to only those in the frustum, but the culling is not based on the center of gaussian location, but the projected 2D bounding box of the Gaussian (based upon the clipped projection here).

Let me know if this makes sense?

Note @f-dy if you naively try your solution to remove the clamping alone, you will get absolutely rubbish results (blury image that looks NOTHING like the scene) because all the Gaussians out to the side are basically filling the whole image with their (incorrect because of linear approximation) projection.
However an alternative (and reasonable, but maybe not as good) solution is to remove all point where their center is outside the 1.3fov (as described in a above)

liruilong940607:
I just finished the backward pass fix as well as the torch impl fix. cc for review @maturk @JonathonLuiten @jb-ye

liruilong940607:
I checked the tests/ locally, all tests passed

Atticuszz:
sorry to bother you guys！thanks for your incredible work of art.may i
 ask a question  about the fixing,does it affect the depth composing of gsplat？

PR #208, opened 2024-06-08
2DGS

FantasticOven2:
This is an attempt to integrate 2DGS [https://surfsplatting.github.io/](url) into gsplat. You can run simple 2D traning example with:
2D verification examples:
`python examples/2d_example.py --model-type {3dgs, 2dgs}
`
Image fitting:
`python examples/image_fitting.py --model-type {3dgs, 2dgs}`

3D scene reconstruction:
`python examples/simple_trained_2dgs.py  --data-dir ...`

https://github.com/user-attachments/assets/310ba772-75e7-4dd3-848a-44e3314d0d08

Evaluations (Average over 7 scenes in Mipnerf-360 dataset)
w/o regularization
| Scene | PSNR | SSIM | LPIPS | Memory (GB) | Train Time (min:sec)| Num GS |
|----------|----------|----------|----------|----------|----------|----------|
| gsplat | 28.76  | 0.8669| 0.1450 | 3.70   | 15:44  | 1.90M |
| gsplat (packed) | 28.74 | 0.8617 | 0.1448 | 3.48 | 15:25 | 1.89M |
| inria | 28.73 | 0.8601 | 0.1480 | 3.73 | 22:16 | 1.93M |

w/ normal consistency regularization
| Scene | PSNR | SSIM | LPIPS | Memory (GB) | Train Time (min:sec)| Num GS |
|----------|----------|----------|----------|----------|----------|----------|
| gsplat | 27.85 | 0.8436 | 0.1853 | 3.78 | 16:48 | 1.96M |
| inria | 28.06 | 0.8492 | 0.1694 | 3.66 | 22:18 | 1.91M

w/ normal consistency, distortion regularization
| Scene | PSNR | SSIM | LPIPS | Memory (GB) | Train Time (min:sec)| Num GS |
|----------|----------|----------|----------|----------|----------|----------|
| gsplat | 27.80 | 0.8422 | 0.1866 | 3.76 | 16:44 | 1.96M |
| inria | 28.05 | 0.8489 | 0.1697 | 3.61 | 22:06| 1.88M |

PR #309, opened 2024-07-31
Compressing gaussian splats

jefequien:
Thanks largely to MCMC's improved densification strategy, a simple post-training compression scheme is enough to top the 3D Gaussian Splatting Compression Methods [leaderboard](https://w-m.github.io/3dgs-compression-survey/). 
This PR implements quantization, sorting, and spherical harmonic coefficients K-means clustering for compression.

**Relevant links:**
https://github.com/fraunhoferhhi/Self-Organizing-Gaussians
https://aras-p.info/blog/2023/09/27/Making-Gaussian-Splats-more-smaller/

## Results

Evaluated on MipNeRF360 dataset.

| Method  | PSNR | SSIM | LPIPS | Size (MB) |
| -------- | ------ | ----- | ------ | ---------- |
| mcmc (0.5M) w/ compression | 28.24  | 0.852 | 0.175 | 8.89 |
| mcmc (1M) w/ compression     | 28.65  | 0.864 | 0.154 | 16.5 |
| mcmc (2M) w/ compression    | 29.02  | 0.872 | 0.138 | 31.2 |

| Method  | PSNR | SSIM | LPIPS | Size (MB) |
| -------- | ------ | ----- | ------ | ---------- |
| mcmc (1M) w/ quant and kmeans     | 28.65  | 0.864 | 0.154 | 16.5 |
| mcmc (1M) w/ quant only | 29.02 | 0.870 | 0.146 | 31.6 |
| mcmc (1M) w/o compression | 29.18 | 0.874 | 0.142 | 236  | 

**Note**: Using K-means centroids and labels to initialize a codebook sets the high score for a given file size, at the cost of another training run. This is not implemented in the PR.

## Data format
This PR quantizes `means` to 16 bits, `opacities`, `quats`, `scales`, `sh0` to 8 bits, and `shN` to 6 bit. Sorting these fields with [PLAS](https://github.com/fraunhoferhhi/PLAS) and saving them as PNGs saves a few MBs and generates nice visuals. 
`means` are quantized in log space.
`shN` coefficients are clustered with K-means into 2**16 clusters and their centroids and labels are saved to a compressed npz file.

Bonsai scene's quantized and sorted `sh0` png.
bonsai_0 5m_sh0


liruilong940607:
This is freaking awesome! At a quick glance, is it missing the handling of the extra features stored in the `splats`, for example, if the splats have a "feature" field or something else, how should these be compressed?

jefequien:
Without metrics, it would be hard to know how much quantization is acceptable. I can add logic that saves any additional field as an npz file with full precision.

liruilong940607:
Oh yes that makes sense. Also should the "meta" be stored together with the compressed data instead of returning it from the function?

jefequien:
All the metadata is stored as a single `meta.json` in the compression directory by the `compress_splats` function.

liruilong940607:
Ah right I missed that! Cool!

gsplat/compression/compress.py
```diff
@@ -0,0 +1,340 @@
...
+from typing import Any
+import json
+import torch
+from torch import Tensor
+import torch.nn.functional as F
+from torchpq.clustering import KMeans
+import numpy as np
+import imageio
+
+from plas import sort_with_plas
```
liruilong940607, line 11: Add an try-except for this import and raise the error with an helper message on how to install it.

liruilong940607, line 11: As of now the only dependency of gsplat is torch. Personally I'm a fun of minimal dependencies for easy maintainess. Since we are not heavily depends on them, maybe it makes sense to push them into `compress_splats` function with an try expect, and add a `.. warning` in the func doc to tell users how to install them.

btw, just curious, if your env does not have them installed, would it fail to run `from gsplat import rasterization`?

jefequien: Added

jefequien: Moved kmeans and plas into their respective functions so you shouldn't run into import errors unless you use the --compress flag.

liruilong940607:
Is there a abbrev. name for this approach? I'm thinking whats the best way to organize the code. If there is some other compression algo coming out in the future, we would have to do a refactor on the naming.

Also it seems like this compression approach does not need any predefined content (e.g., a network), which is nice. But what if a future one uses some kind of network, or any other predefined content? You must know more than me on this front.

Ideally we don't have to refactor the code once another compression method is PR'd. So maybe it makes sense to make it into a class as well? Like the `Strategy`?

jefequien:
There isn't a name but the sorting is from Self Organizing Gaussians and the K-means idea is from the blog series. That makes sense. I could refactor this into a class to leave room for a different compression strategy.

jb-ye:
Why is MCMC better than default method in terms of k-means based compression? @jefequien

jefequien:
Both methods lose about ~0.5 PSNR after quantization and k-means. MCMC just gives a higher starting point than the default settings.

Ben-Mack:
@jefequien  I'm having this error while trying this PR, do you have any idea?
```
python M:\gsplat-compress\examples\simple_trainer.py --data_dir ./ --result_dir ./res --data_factor 1 --compression_strategy png

File "C:\Users\admin\miniconda3\envs\gscompress\lib\site-packages\gsplat\compression\sort.py", line 5, in <module>
    def sort_splats(splats: dict[str, Tensor], verbose: bool = True) -> dict[str, Tensor]:
TypeError: 'type' object is not subscriptable
```

I already install gsplat of this PR with python 3.8 following similar to this guide https://github.com/nerfstudio-project/gsplat/pull/208#issuecomment-2273996152

liruilong940607:
I think all `dict` needs to be `Dict` imported from `typing`.

I'll take over this PR and do some cleanup in the coming days

jefequien:
@Ben-Mack Right. I personally use python 3.10 so as a temporary fix, you can try upgrading your python version.

brentyi:
> I think all `dict` needs to be `Dict` imported from `typing`.

As an FYI `from __future__ import annotations` at the top of files will also let you use `dict[K, V]`, `tuple[T]`, etc!

liruilong940607:
Did some cleanup and put up a doc page for it:

Screenshot 2024-08-12 at 11 39 25 AM

tests/test_compression.py
```diff
@@ -0,0 +1,44 @@
...
+    # Prepare Gaussians
+    N = 10000
+    splats = torch.nn.ParameterDict(
+        {
+            "means": torch.randn(N, 3),
+            "scales": torch.randn(N, 3),
+            "quats": torch.randn(N, 4),
+            "opacities": torch.randn(N),
+            "sh0": torch.randn(N, 1, 3),
+            "sh1": torch.randn(N, 24, 3),
```
jefequien, line 30: Typo. shN?

liruilong940607: fixed!

liruilong940607 (approved):

Florian-Barthel:
Great to see [Self-Organizing-Gaussians](https://github.com/fraunhoferhhi/Self-Organizing-Gaussians) beeing used!

SharkWipf:
FYI: I was messing around with the code with Claude, and it seems to inadvertently have caught and fixed a bug related to the compression in simple_trainer.py:

image

It seems correct to me at a glance, but as I am not read in into the code I don't want to open an issue/PR for an accidental Claude change that I haven't looked into, so I figured I'd just put it here instead so you can judge it yourself.

ichsan2895:
Hello, sorry for late party. I have used this method in `gsplat==1.3.0`. But I get the compression format is PNG instead of PLY.

I must decompress it, so I can convert it to PLY. But, unfortunatelly, the size of PLY is remain same.

The question is, any tutorial for getting compressed PLY with this method?

ichsan2895:
> FYI: I was messing around with the code with Claude, and it seems to inadvertently have caught and fixed a bug related to the compression in simple_trainer.py:
> 
> image
> 
> It seems correct to me at a glance, but as I am not read in into the code I don't want to open an issue/PR for an accidental Claude change that I haven't looked into, I figured I'd just put it here instead so you can judge it yourself.

Yeah, I got this error too. It should be
```
compress_dir = f"{self.cfg.result_dir}/compression/rank{world_rank}"
```
It does not error somehow in my setup with 1 GPU, but if it has more >1 GPU, it caused error

Ben-Mack:
Hi @jefequien, are there anyway to load Checkpoint from Nerfstudio and run just the compression step, instead of training from ground up in gsplat?

jefequien:
@Ben-Mack Assuming Nerfstudio uses the same .pt file as gsplat, it should be simple. Just use simple_trainer.py with --ckpt
https://github.com/nerfstudio-project/gsplat/blob/cc2fb4c5690fe9e70a086214001fafa0c9f06a4c/examples/benchmarks/compression/mcmc.sh#L39

Issue #318, opened 2024-08-07
JIT-compiled extension and library version.


keunhong:
It seems like there's no mechanism to detect whether the library has been updated in the code that checks whether the compiled extensions exist.
https://github.com/nerfstudio-project/gsplat/blob/7816033ef1ec0b1fe890cb251412efde19e37f36/gsplat/cuda/_backend.py#L105

This means the compiled extension stays stale even after the library is updated with `pip`.

I think an easy fix is to just add the library version to the name in: https://github.com/nerfstudio-project/gsplat/blob/7816033ef1ec0b1fe890cb251412efde19e37f36/gsplat/cuda/_backend.py#L83

PR #320, opened 2024-08-07
Add gsplat version number to build directory.

keunhong:
This ensures that the extension is compiled when the library is upgraded.

Resolves https://github.com/nerfstudio-project/gsplat/issues/318.

### DIFF
--- a/gsplat/cuda/_backend.py
+++ b/gsplat/cuda/_backend.py
@@ -11,6 +11,9 @@ from torch.utils.cpp_extension import (
     load,
 )
 
+import gsplat
+
+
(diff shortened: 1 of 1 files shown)


liruilong940607:
I did a test on this. It seems like at least in the latest release it will trigger the rebuilt if you upgrade it with pip, which is as expected. So I'm closing this for now. Feel free to reopen it if this is not the case.

```
# install v1.1.0
pip install gsplat==1.1.0
python -c "from gsplat.cuda._backend import _C; print(_C)"

# replace v1.1.0 with v1.3.0
pip install gsplat==1.3.0
python -c "from gsplat.cuda._backend import _C; print(_C)"
```

PR #321, opened 2024-08-08
Fix typo in examples/simple_viewer.py

simo23:


maturk (approved):
LGTM thanks for the fix!

PR #325, opened 2024-08-10
Consolidate simple_trainer.py and simple_trainer_mcmc.py

brentyi:
Not fully tested yet, but here's a config structure suggestion to make this API possible: `python simple_trainer.py {default,mcmc}`.

`python simple_trainer.py --help`:

image

**Default densification:** `python simple_trainer.py default --help`

image

**MCMC densification:** `python simple_trainer.py mcmc --help`

image


brentyi:
Pushing us in the direction of nerfstudio (`splatfacto` vs `splatfacto-big`, etc), this also makes it easier to have different sets of standard hyperparameters.

You can just add a new key/value pair to this dictionary:
```python
    # Config objects we can choose between.
    # Each is a tuple of (CLI description, config object).
    configs = {
        "default": (
            "Gaussian splatting training using densification heuristics from the original paper.",
            Config(),
        ),
        "mcmc": (
            "Gaussian splatting training using densification  from the paper '3D Gaussian Splatting as Markov Chain Monte Carlo'.",
            Config(init_opa=0.5, init_scale=0.1, strategy=MCMCStrategy()),
        ),
    }
```

liruilong940607 (approved):
LGTM! Thanks Brent!

PR #332, opened 2024-08-15
Simplify `tyro` usage in `simple_trainer.py`

brentyi:
The subcommand pattern we were using is really common, but required some magic. I moved this to a helper inside of `tyro`. (unfortunately this requires upgrading)


liruilong940607 (approved):
LGTM!

PR #334, opened 2024-08-16
Patch to make Strategy compatible with splatfacto

kerrj:
This is because nerfstudio uses (N,1) for opacity, and gsplat uses (N,).
I don't think this should affect gsplat functionality, but I haven't tested so if someone else could that'd be great!


liruilong940607:
Right! Sorry that I forgot about this! 

I changed the `squeeze` to `flatten` to avoid the shape error in the extreme case where there is only a single gaussian.

Also modify a few more places.

This seems to be good but not tested.

liruilong940607 (approved):

PR #349, opened 2024-08-21
Added orthographic projection

VladislavZavadskyy:
Should be useful :)


gsplat/cuda/_torch_impl.py
```diff
@@ -106,6 +106,43 @@ def _persp_proj(
...
 
 
+def _ortho_proj(
+    means: Tensor,  # [C, N, 3]
+    covars: Tensor,  # [C, N, 3, 3]
+    Ks: Tensor,  # [C, 3, 3]
+    width: int,
+    height: int,
+) -> Tuple[Tensor, Tensor]:
+    """PyTorch implementation of prespective projection for 3D Gaussians.
```
liruilong940607, line 116: typo: "perspective projection"

gsplat/cuda/_wrapper.py
```diff
@@ -77,12 +77,13 @@ def quat_scale_to_covar_preci(
     return covars if compute_covar else None, precis if compute_preci else None
 
 
-def persp_proj(
+def proj(
```
liruilong940607, line 80: For the sake of backward compatibility, I think we should keep the function `persp_proj` and have another `ortho_proj`. The rest of the code does not need to change (i.e., these two APIs could share the underlying implementation)

Another solution which is also fine with me, is to have this `proj` function and keep the original `persp_proj` but raise deprecated warning inside of it. This would also avoid sudden interrupt to the users.

tests/test_basic.py
```diff
@@ -174,6 +180,8 @@ def test_projection(test_data, fused: bool, calc_compensations: bool):
     quats = test_data["quats"]
     scales = test_data["scales"]
     means = test_data["means"]
+    ortho = test_data.get("ortho", False)
```
liruilong940607, line 183: You can set `ortho` as an argument of this function and decorate this function with `pytest` to specify this argument to be with options True and False. This way pytest would test over all choices. See examples here: https://github.com/nerfstudio-project/gsplat/blob/45d196a3611e627c498f237c27014cdaf42d94d2/tests/test_rasterization.py#L21-L23

liruilong940607 (changes requested):
This is very useful!! It seems like a handful of users are looking for this support from #349! Left a few minor comments.

liruilong940607:
Test is failing because of format. Could you run
```
pip install black==22.3.0
black . gsplat/ tests/ examples/ profiling/
```

liruilong940607 (approved):
LGTM! Thanks!

VladislavZavadskyy:
Pleasure doing business with you :)

szhang963:
Hi, @VladislavZavadskyy 
Thanks for the contribution. 
I found render quality degradation (like, more blurry, more artifacts) in orthographic projection compare to perspective projection for a plane (e.g. ground). 
Is it normal phenomenon? 
If not ,could you provide some suggestions?
Thanks for any help.

VladislavZavadskyy:
Hi, @szhang963. Yeah, I'd say it's expected, since the Gaussians were fitted using first order approximation of perspective projection, which is often incorrect, parameters take that incorrectness into account so other projection modes may produce results which are not as good. If that's critical for you, you might try some of the perspective-correct training methods.

VladislavZavadskyy:
Also, keep in mind, that gaussians by themselves have no geometry prior, so the result you'll get is very dependent on the quantity and quality of training views. I mean that even if you use perspective-correct methods, but you have limited number of views of some object, for example, you'll most likely see bad results outside of the training views + some neighborhood, regardless of the projection. Different projection may make it more obvious though. 

There shouldn't be any issues in the implementation, since it plain and simple, but feel free to look for yourself.

szhang963:
Thanks for your reply, @VladislavZavadskyy.
Could you provide me some references for perspective-correct methods? Or, how can I remain the render quality of perspective projection under orthographic projection?

szhang963:
Hi @VladislavZavadskyy, I have solved this issue. I found it is caused by viewdir in spherical harmonics. Specifically, I did not set correct position of orthographic camrea. In fact, the quality of orthographic rendering is not poor compare to perspective rendering.

gwen233666:
> 你好@VladislavZavadskyy我已经解决了这个问题。我发现这是由球谐函数中的 viewdir 引起的。具体来说，我没有设置正确的正交相机位置。事实上，正交渲染的质量与透视渲染相比并不差。

Hello, I hope you're doing well. Would you be so kind as to provide me with some guidance or explain how you implemented orthographic rendering? I would greatly appreciate your help. Thank you!

ccl-1:
> Hi @VladislavZavadskyy, I have solved this issue. I found it is caused by viewdir in spherical harmonics. Specifically, I did not set correct position of orthographic camrea. In fact, the quality of orthographic rendering is not poor compare to perspective rendering. 

Hi, @szhang963  what you mean "caused by viewdir in spherical harmonics"? how to set correct position of orthographic camrea?  My render quality under orthographic projection is not good.

VladislavZavadskyy:
@ccl-1 You'd need to provide the same ray direction for all harmonics which coincides with camera Z-axis.

ccl-1:
Hi @VladislavZavadskyy , i use the interface of 3DGS. I modified the default view matrix (equal=perspective matrix, originally aligned with the car's driving direction) to make it coincides with camera Z-axis.. The render quality under orthographic projection looks stranfe. And I can't find where to set the ray direction for harmonics.  The code is as follows. Could you confirm if this is the correct setup? Many thanks!

```
from gsplat import rasterization
def render(viewpoint_camera, pc : GaussianModel, pipe, bg_color : torch.Tensor, scaling_modifier = 1.0, 
           separate_sh = False, override_color = None, use_trained_exp=False):
 
    # Set up rasterization configuration
    tanfovx = math.tan(viewpoint_camera.FoVx * 0.5)
    tanfovy = math.tan(viewpoint_camera.FoVy * 0.5)
    
    focal_length_x = viewpoint_camera.image_width / (2 * tanfovx)
    focal_length_y = viewpoint_camera.image_height / (2 * tanfovy)
    
    K = torch.tensor(
        [
            [focal_length_x, 0, viewpoint_camera.image_width / 2.0],
            [0, focal_length_y, viewpoint_camera.image_height / 2.0],
            [0, 0, 1],
        ],
        device="cuda",
    )

    viewmat = viewpoint_camera.world_view_transform.transpose(0, 1) # perspective

    width=int(viewpoint_camera.image_width),
    height=int(viewpoint_camera.image_height),

    if camera_model =='ortho':
        width=1200
        height=1500

        theta = math.pi / 2  # forward view -> top view
        Rx = torch.tensor([
            [1, 0, 0],
            [0, math.cos(theta), -math.sin(theta)],
            [0, math.sin(theta), math.cos(theta)]
        ], dtype=torch.float32).cuda()

        R_bev = Rx @ viewmat[:3, :3]
        viewmat[:3, :3] = R_bev
        viewmat[1, 3] +=2 
        viewmat[2, 3] +=1 # camera hight

    means3D = pc.get_xyz
    opacity = pc.get_opacity

    scales = pc.get_scaling * scaling_modifier
    rotations = pc.get_rotation

    if override_color is not None:
        colors = override_color # [N, 3]
        sh_degree = None
    else:
        colors = pc.get_features # [N, K, 3]
        sh_degree = pc.active_sh_degree
    
    render_colors, render_alphas, info = rasterization(
        means=means3D,  # [N, 3]
        quats=rotations,  # [N, 4]
        scales=scales,  # [N, 3]
        opacities=opacity.squeeze(-1),  # [N,]
        colors=colors,
        viewmats=viewmat[None],  # [1, 4, 4]
        Ks=K[None],  # [1, 3, 3]
        backgrounds=bg_color[None],
        width=width,
        height=height,
        packed=False,
        sh_degree=sh_degree,
        camera_model=camera_model
    )```
![res](https://github.com/user-attachments/assets/9b128643-d53a-44e8-9d58-20bf4837922e)

PR #351, opened 2024-08-21
automatically add the micosoft visual studio compiler binary to the environment path

martinResearch:
In some case gsplat needs to compile cuda/c++ code on the fly (just-in-time compilation), 
to handle different number of color channels for example, like done in the test `test_rasterize_to_pixels` when using `channels=128`. 

In that case `cl.exe` needs to be on the path when on windows in order for pytorch to be able to do the compilation.
One can follow the step documented [here](https://github.com/nerfstudio-project/gsplat/blob/main/docs/INSTALL_WIN.md)
but that requires a manual step each time time one wants to use gsplat, which is not convenient.

We can instead add the folder automatically to the path at runtime using a mechanism used in Cupy to discover the folder that contains the `cl.exe` file.

Ideally this mechanism would be implemented in pytorch, but we can implement it in gsplat in the mean time.

### DIFF
--- a/gsplat/cuda/_backend.py
+++ b/gsplat/cuda/_backend.py
@@ -10,6 +10,7 @@ from torch.utils.cpp_extension import (
     _import_module_from_library,
     load,
 )
+import sys, warnings
 
 PATH = os.path.dirname(os.path.abspath(__file__))
 NO_FAST_MATH = os.getenv("NO_FAST_MATH", "0") == "1"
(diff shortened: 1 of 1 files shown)


liruilong940607:
Hey this looks useful! I'm not a windows player but this PR in general looks reasonable to me, with a few question:

> In some case gsplat needs to compile cuda/c++ code on the fly (just-in-time compilation),
to handle different number of color channels for example, like done in the test test_rasterize_to_pixels when using channels=128.

Just want to clarify `channels=128` should not be any special w.r.t. compiling gsplat. If you install gsplat via `pip install git+https://github.com/nerfstudio-project/gsplat/`, it will trigger `setup.py`, which will build (compile) at installation and it should not trigger JIT after installation in any cases.

> In that case cl.exe needs to be on the path

What is `cl.exe`? Is it something needed for JIT compile only?

martinResearch:
Cl.exe is the executable used to run the Microsoft visual studio compiler and linker. I misinterpreted the need for cl when running some of the tests as being due to jit compiling still being required after the package Install. I will investigate why cl is still needed for some tests at runtime.

martinResearch:
I realized the problem is in the `nerfacc` library, not in `gsplat`.
The installation using `pip install nerfacc` uses the pipy.org file `nerfacc-0.5.3-py3-none-any.whl` and not `nerfacc-0.5.3.tar.gz` and as a results the CUDA code is not compiled during the installation, but compiled one the fly the first time `nerfacc` is imported.
On windows, this requires `cl.exe` to be on the path at runtime. A workaround is to install nerfacc from source using  `pip install --no-binary=nerfacc  nerfacc`

Issue #363, opened 2024-08-22
torch missing when running setup.py


martinResearch:
**Problem**
I am trying to install gsplat from source among other librairies with a `requirements.txt` and `pip install -r requirements` but that fails because it tries to compile gsplat with pytorch before pytorch is installed (all the package get installed at the end after all the packages are build).

**Solution**
To prevent this, pytorch should be listed in a list provided of packages needed to run `setup.py` through the `setup_requires` argument in setup.py (see https://setuptools.pypa.io/en/latest/references/keywords.html) or using https://peps.python.org/pep-0518/

liruilong940607:
Hey! It's actually pretty tricky to put pytorch into `setup.py` so we rely on user to install it manually before installing gsplat.

Installation of pytorch depends on the CUDA version that lives in your machine, and different CUDA corresponds to different installation command, as shown here: https://pytorch.org/get-started/locally/

The fundamental reason is that the [platform compatibility tags](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/) supported by PyPI does not include CUDA version. So PyPI can not be used to host wheels for different CUDA versions (and thats why PyTorch was hosting them on their own server and ask user to install via `--index-url https://download.pytorch.org/whl/cuXXX`)

gsplat also depends on CUDA, and has one more dependence, which is PyTorch. Neither of them can be recognized by PyPI tages so if we build wheels we would have to host it somewhere else, just like PyTorch.

martinResearch:
Closing as there is not obvious solution to this problem. The precompiled wheels are now available (https://github.com/nerfstudio-project/gsplat/pull/365) which provides a work around.

PR #364, opened 2024-08-23
Add pyproject.toml build system

martinResearch:
trying to address issue #363


pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+# Minimum requirements for the build system to execute.
+requires = ["setuptools", "wheel", "torch", "numpy==1.26.4"]  # PEP 508 specifications.
```
liruilong940607, line 3: I guess `pip install .` would trigger this, is that true? Would this get triggered if I run `python setup.py`? 

Also `numpy==1.26.4` seems too constrained than necessary. We probably don't need to specify numpy version for the gsplat library?

liruilong940607:
Btw, I think the requirements are already there: https://github.com/nerfstudio-project/gsplat/blob/507f26e118afb78ef0224ad5e0d0701f1b973853/setup.py#L130-L137

PR #365, opened 2024-08-23
Publish precompiled wheels in Github release and add wheel index

martinResearch:
The PR adds a mechanism to publish precompiled wheels in github releases and pypi and add a wheel index html file to be able to install precompiled wheels for a specific pytorch and cuda version using a single line like

```
pip install gsplat --index-url https://martinresearch.github.io/gsplat/whl/pt20cu118
```
One can also omit the `pt20cu118` and use 
```
pip install gsplat==1.2.0+pt20cu118 --index-url https://martinresearch.github.io/gsplat/whl
```
but then the full version needs to be specified to make sure one gets the right pytorch and cuda version.

The full list of pre-compiled wheel can be found [here]( https://martinresearch.github.io/gsplat/whl/gsplat)
Is that a solution you would potentially be happy to merge in your repository?

This PR also tries to upload all the wheels to pypi I am not sure  pipy will allows us to upload wheels with "+" sign in the wheel name (i.e using local version to specify the cuda and pytorch versions), and I am not sure whe want to upload these wheels to pypi, so the upload step could be removed for the precompiled wheels


liruilong940607:
Hey I like this way a lot! I could give it a try during the weekend. Do you know if GitHub release has storage limit for it?

martinResearch:
@liruilong940607 not sure about the size limit. Can't find in the Github doc.
Note that the project https://github.com/yt-dlp/yt-dlp also published precompiled wheels in the release. Each of their release seem to take at least 50MB and they have about 90 releases. I don't now if they have a free account and if github has different size limits for free accounts.

whl/gsplat/index.html
liruilong940607, line 1: How are these html files generated? And how should we maintain these files? Actually Is there a way to avoid these files living in the gsplat repo? After we have multiple releases we would have a huge amount of folders and html files if we put them in the gsplat repo.

liruilong940607:
I took a closer look. This is a nice way but I think there is a few drawbacks that would be best if we can avoid them.

I also updated the `publish.yml` in #369 to support a complete permutations of all version, which are all tested to be buildable. 

Ideally we should build all these permutations at each release but putting all these wheels as github release assets feels a bit heavy to me (not unacceptable). Or we drop some permutations and only provide wheels for limited version combinations, which are also fine with me.

.github/workflows/building.yml
```diff
@@ -15,11 +15,14 @@ jobs:
...
         # support version based on: https://download.pytorch.org/whl/torch/
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
-        torch-version: [1.11.0, 1.12.0, 1.13.0, 2.0.0]
-        cuda-version: ['cu113', 'cu115', 'cu116', 'cu117', 'cu118']
+        #python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        #torch-version: [1.11.0, 1.12.0, 1.13.0, 2.0.0]
+        #cuda-version: ['cu113', 'cu115', 'cu116', 'cu117', 'cu118']
+        python-version: ['3.10',]
+        torch-version: [2.0.0, ]
+        cuda-version: ['cu118',]
```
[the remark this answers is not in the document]
martinResearch: Sorry I did not mean to keep these changes and will revert them. I agree we should ideally test all combination, but we probably do not need to publish all of them in GitHub release.  it is probably ok to drop python 3.7 and 3.8  in the published wheels as these a fairly only version. Python 3.7 that does doe get security fixes (end of life phase) since June 2023, and python 3.8 will stop receiving security fix after October this year. I would also think that very few people still use pytorch 1.11.0 and either use 1.13.0 or 2.x

.github/workflows/building.yml
```diff
@@ -77,6 +80,20 @@ jobs:
         with:
           submodules: 'recursive'
 
+      - name: Free up disk space
```
martinResearch, line 83: @liruilong940607 for some reason I had to move this step earlier in the pipelines and delete more folder for it to run in github action. We may want to test the piepline in the main branch first and we could do that change in a separate PR to get it working

.github/workflows/building.yml
```diff
@@ -129,9 +142,13 @@ jobs:
           BUILD_NO_CUDA=1 pip install .
 
       - name: Build wheel
-        run: |
+        # architectures 8.x and 9.x (Ampere, Ada, Hopper) are not avaible in the current github action
```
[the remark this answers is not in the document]
martinResearch: indeed I can build for 8.0;8.6;9.0. I think it did not work for 8.9 and 8.9, but I need to double check. If that is the case we need to make it clear in the doc and recommend installation from source when people have gpu with these architecture if that makes a significative difference in speed.

.github/workflows/publish.yml
```diff
@@ -1,36 +1,240 @@
...
 # For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
 
-name: Upload Python Package
+name: Build and Release Wheels
 
-on:
-  release:
-    types: [created]
-    branches: [main]
+on: [push]
```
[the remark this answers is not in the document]
martinResearch: I agree, did this change while doing tests, will revert.

.github/workflows/publish.yml
```diff
@@ -1,36 +1,240 @@
...
+    - uses: actions/download-artifact@v2
+      with:
+        name: packages
+        path: dist
+
+    - name: Publish package to Test PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        repository-url: https://test.pypi.org/legacy/
```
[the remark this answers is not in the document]
martinResearch: est PyPI allows developers to test the process of uploading their Python packages before they make them available on the official PyPI repository. This helps ensure that the packaging, metadata, and distribution are set up correctly.

whl/gsplat/index.html
martinResearch: I created them manually. We could generate them automatically with a python script but that seems a bit of an overkill. I am not sure there is a way to avoid having them in the repo, but we can keep them in a different branch than the main branch to avoid cluttering the main branch. It seems to me the number of folder should remain manageable for quite a while. if we support only cu117 and cu118 only 2 folder will need to be added when pytorch 2.5 is released.

martinResearch:
> I took a closer look. This is a nice way but I think there is a few drawbacks that would be best if we can avoid them.
What drawbacks do you have in mind? I believe we can address all of the comments you made on this PR

.github/workflows/publish.yml
```diff
@@ -1,36 +1,240 @@
...
+        release_name: Release ${{ env.VERSION }}
+        body: |
+          Release notes for version ${{ env.VERSION }}.
+        draft: false
+        prerelease: false
+
+    - name: Download packages
+      id: download_artifacts
+      uses: actions/download-artifact@v3
+      with:
```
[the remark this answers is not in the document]
liruilong940607: Yeah this is by far my biggest question. Thanks for looking into it!

whl/gsplat/index.html
liruilong940607: I think torch is gradually shifting to cuda 12 -- e.g., torch 2.4.0 supports cuda 118, 121 and 124.  So we would need roughly 3x more folders than what we have now. But I would be happy if we move then to another branch (maybe merge them with the web page branch `gh-pages` and  use a script to generate them?). Because then one pull the main branch would not pull these files.

liruilong940607:
> > I took a closer look. This is a nice way but I think there is a few drawbacks that would be best if we can avoid them.
> 
> What drawbacks do you have in mind? I believe we can address all of the comments you made on this PR

The comments are all the concerns/questions I have! If we could address them I would be very happy to see this get merged!

.github/workflows/publish.yml
```diff
@@ -1,36 +1,240 @@
...
+            cuda-version: cu118
+
+          - os: windows-2019
+            python-version: '3.10'
+            torch-version: 2.2.2
+            cuda-version: cu118
+
+          - os: windows-2019
+            python-version: '3.10'
+            torch-version: 2.3.1
```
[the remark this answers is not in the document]
martinResearch: I removed duplication by having the publish workflow calling call the build workflow

whl/gsplat/index.html
martinResearch: yes we can have the html pages in gh-pages. I am not sure why having them in main would be a problem though. These are small files and they are all under the whl folder

.github/workflows/building.yml
martinResearch: actually seems to work now. Not sure why I have been hitting a problem before

.github/workflows/building.yml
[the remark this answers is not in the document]
martinResearch: for torch audio , each torchaudio version is compatible with only one torch version (see compatibility matrix at the bottom of the page here)
https://pytorch.org/audio/2.1/installation.html
so that seems to indicate that different torch 2.x version do not have compatible ABI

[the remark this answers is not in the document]
liruilong940607: Yeah I also think we could drop 1.13. 

How about lets go with three latest python version, three latest torch versions and three latest CUDA version. Then we could build against:

- python 3.10, 3.11, 3.12
- torch 2.2, 2.3, 2.4
- cuda 118 121 124

I'm not a fan of torchaudio's strategy. Their strategy is ok only if the library itself is not quite updating that much, and it's just about to bump the dependency version.

whl/gsplat/index.html
liruilong940607: I think the ideal workflow is that when we create a release, the building workflow is triggered, in which there will be a script to generate these html files based on the versions we are compiling against. This is ideally to be automatic because in the future when torch 2.5 came out we could only update the workflow. That being said if the script that generates the htmls push these files back to main it will create a new commit, which is a bit weird to happen right after the release. And since you packaged main into a .tar.gz, this release file will include those out-dated htmls in it.

So if we instead pushing these htmls to another branch we would not have the above issues. Philosophically these htmls share the same property with the web pages, which need to be updated together with release. That's why I thought putting them into the `gh-pages` branch might be better.

Plus, it would be really nice if we could pin the`--index-url` to `https://docs.gsplat.studio/`, like
```
pip install gsplat --index-url https://docs.gsplat.studio/whl/pt20cu118
```

bolopenguin:
That would be great! Any news?

martinResearch:
> That would be great! Any news?

I will try to finish addressing
The comment on automatic index.html files generation this weekend.

whl/gsplat/index.html
martinResearch: I added a step to generate the html pages automatically in the gh-pages branch

martinResearch:
@liruilong940607 thank you very much for the feedback. I tried to address your comments. Any remaining concern or improvements suggestion before I publish the PR?

.github/workflows/generate_simple_index_pages.py
```diff
@@ -0,0 +1,152 @@
...
+        print("No Python wheels found in the releases.")
+
+    # Generate Simple Index HTML pages the wheel with all local versions
+    generate_simple_index_htmls(wheels, outdir=args.outdir)
+
+    # group wheels per local version
+    wheels_per_local_version = {}
+    for wheel in wheels:
+        local_version = wheel["local_version"]
+        if local_version not in wheels_per_local_version:
```
martinResearch, line 132: if local_version is not and and local_version not in wheels_per_local_version:

liruilong940607:
This looks great! Thanks for all the updates! I'm doing test now

liruilong940607 (approved):
I tested it out and did some fixes here and there. Now I think it is good to go. Thanks a lot for digging into this @martinResearch !

PR #385, opened 2024-09-04
Add mipnerf360_extra, tandt, deepblending to download_dataset.py

paulbzm:
This adds 2 new datasets for evaluation to the download script, as well as the 2 extra scenes from mipnerf360.
Datasets can be selected with --dataset, select "all" to download all datasets.

PR #291, opened 2024-07-17
Implement Bilateral Guided Gaussian Splatting.

jefequien:
Bilateral Guided Radiance Fields showed that bilateral grids are very effective at modeling exposure changes and other ISP effects. This PR ports their bilateral grid library onto Gaussian splatting. It seems quite effective at removing floaters in casually captured videos and improves color-corrected PSNR on the bilarf dataset.

https://github.com/user-attachments/assets/2f97afc4-c136-4e6f-9fe4-d8f1a7bc4ed0

https://github.com/user-attachments/assets/2d7942ea-49dc-41f8-ad80-0a11a2b30645

Original authors also expected this to work quite well. https://github.com/yuehaowang/bilarf/issues/1

TODO:
- [x] Match original paper's learning rate schedule exactly.

Does this belong in gsplat or is this more appropriate for nerfstudio?

Issue #252, opened 2024-07-02
Haze with MCMC with drjohnson dataset


ichsan2895:
### Haze or soft floaters does not appear in gsplat-Default implementation with `drjohnson` dataset from Deep Blending. But it has lower metrics than gsplat-MCMC
```
python3 gsplat/examples/simple_trainer.py --eval_steps -1 --disable_viewer --data_factor 1 \
        --data_dir data/drjohnson_PLY \
        --result_dir output/drjohnson_PLY

python3 gsplat/examples/simple_trainer.py --disable_viewer --data_factor 1 \
            --data_dir data/drjohnson_PLY \
            --result_dir output/drjohnson_PLY \
            --ckpt output/drjohnson_PLY/ckpts/ckpt_29999.pt

python3 gsplat/examples/simple_viewer_unbounded.py --port 7007 \
    --ckpt output/drjohnson_PLY/ckpts/ckpt_29999.pt --backend gsplat
```
PSNR: 29.322, SSIM: 0.9136, LPIPS: 0.159 Time: 0.033s/image Number of GS: 3191483

![image](https://github.com/nerfstudio-project/gsplat/assets/60387342/f8ec9d09-eb58-42d2-9098-e655659f81e6)

### Haze or soft floaters appears in gsplat-MCMC implementation with `drjohnson` dataset from Deep Blending. But it has higher metrics than gsplat-default
```
python3 gsplat/examples/simple_trainer_mcmc.py --eval_steps -1 --disable_viewer --data_factor 1 \
        --data_dir data/drjohnson_PLY \
        --result_dir output/drjohnson_PLY_MCMC

python3 gsplat/examples/simple_trainer_mcmc.py --disable_viewer --data_factor 1 \
            --data_dir data/drjohnson_PLY \
            --result_dir output/drjohnson_PLY_MCMC \
            --ckpt output/drjohnson_PLY_MCMC/ckpts/ckpt_29999.pt

python3 gsplat/examples/simple_viewer_unbounded.py --port 7007 \
    --ckpt output/drjohnson_PLY_MCMC/ckpts/ckpt_29999.pt --backend gsplat
```

PSNR: 29.788, SSIM: 0.9202, LPIPS: 0.168 Time: 0.035s/image Number of GS: 1000000

![image](https://github.com/nerfstudio-project/gsplat/assets/60387342/f023c6d6-6901-4eec-81c2-25fc34e5b6b9)

Additional info:
I do not know how to rotate it

System Info:
[...]

PR #388, opened 2024-09-05
Add Deep Blending results, full evaluation script, and per-scene results

paulbzm:
This adds a new evaluation script for the Deep Blending dataset, a new evaluation script to eval all datasets with different numbers of Gaussians and write the per scene results in csvs for the 3dgs.zip survey.
To this end the 2 existing evaluation scripts are modified to accept arguments.

The PSNR results on the Deep Blending dataset (dark brown) seem strange as they decrease with more Gaussians, but the SSIM and LPIPS results seem plausible. Any ideas on how to explain this or if it could be a bug related to the data loading are welcome @jefequien.

Capture
Capture2
Capture3

The dataset can be downloaded with the updated download script in https://github.com/nerfstudio-project/gsplat/pull/385.

When skipping through the renderings, compress_0022.png of scene playroom (1M Gaussians) was the only really bad results that i noticed

compress_0022

However this is probably only an outlier due to lightening because it looks very different in compress_0018.png in a slightly different angle. 

compress_0018

So maybe it is caused by the data itself that this approach leads to decreasing PSNR results with more Gaussians..


liruilong940607:
Thanks for the efforts!

I never played with this dataset, but from the images you posted it seems like at least for this scene the exposure changes need to be compensated. The `simple_trainer.py` has an option `app_opt` to optimize for an appearance embedding for it but that is by default not being compressed I believed. The Bilateral Grid PR #291 is also another nice solution for this exposure issue which I think is more compatible with the compression.

BTW, if I would to guess what is happening for the really bad image, I think there is a huge transparent Gaussian living in the scene, which overfits the exposure in the training images but mess up the rendering from this viewpoint.

jefequien:
@paulbzm Issue #252 reports poor performance on the playroom scene under default mcmc settings. One of the original author recommends, per the paper, setting opacity_reg to 0.001.

PR #396, opened 2024-09-10
Fused differentiable SSIM

rahul-goel:
I've made my Fused SSIM (from Taming 3DGS) implementation public: https://github.com/rahul-goel/fused-ssim. It is really beneficial to use this when the number of gaussians is less since SSIM has a major overhead. Currently, the pixel permutation will also take a lot of time but it is how the rasterizer is implemented and it will require a big re-work so I'm avoiding it.

I've tested the simple trainer on till 7k iterations on the garden scene from Mip-NeRF360 dataset. Using fused-ssim brings down the training time from 5:52 to 4:54 on my RTX 3080Ti.

I've added it as a submodule. But the instructions to install this submodule have not been added yet. Please let me know what exactly to do for it.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "gsplat/cuda/csrc/third_party/glm"]
 	path = gsplat/cuda/csrc/third_party/glm
 	url = https://github.com/g-truc/glm.git
+[submodule "third_party/fused-ssim"]
+	path = third_party/fused-ssim
+	url = git@github.com:rahul-goel/fused-ssim.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
 [submodule "gsplat/cuda/csrc/third_party/glm"]
 	path = gsplat/cuda/csrc/third_party/glm
 	url = https://github.com/g-truc/glm.git
-[submodule "third_party/fused-ssim"]
-	path = third_party/fused-ssim
-	url = git@github.com:rahul-goel/fused-ssim.git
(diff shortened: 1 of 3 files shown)


liruilong940607:
Hey Rahul! This is SUPER cool! And congrats on the release of the `fused-ssim`!

For using it in the `examples/simple_trainer.py`, I think a easier way is probably just adding it as an dependency in the `examples/requirements.txt`:

```
git+https://github.com/rahul-goel/fused-ssim
```

Or if you want to pin on a specific commit in case you might break API in the future:
```
git+https://github.com/rahul-goel/fused-ssim@84422e0da94c516220eb3acedb907e68809e9e01
```

rahul-goel:
Hi.

Thanks for the suggestion on how to deal with it. I've updated the code.

FindDefinition:
pixel permutation in Taming 3DGS is actually greatly slower than simple warp reduce, no need to waste time on it.
another simple way to increase train performance is just add `fused=True` to `Adam` (~20% faster on garden 7k). for sparse Adam, we need to implement a Cuda kernel to increase performance.

rahul-goel:
Hello @FindDefinition.

I think you misunderstood my comment. I mean to say that the permutation function will slow things down while taking the loss:

```
ssimloss = 1.0 - self.ssim(pixels.permute(0, 3, 1, 2), colors.permute(0, 3, 1, 2))
```

But since the rasterizer gives out pixels in such an order, not much can be done.

liruilong940607 (approved):
LGTM!

PR #397, opened 2024-09-10
Make download script work on Windows, too

cr333:
1. Recent versions of Windows 10 don't have `wget` but do have `curl`, so now we can download files across the major platforms.
2. Windows 10 also doesn't have the `unzip` program, but a version of `tar` (apparently bsdtar), which supports `.zip` files (unlike the GNU tar on the Linux box I tested on). So need to use different code paths depending on platform.

Tested on Windows 10 22H2 and Linux.

### DIFF
--- a/examples/datasets/download_dataset.py
+++ b/examples/datasets/download_dataset.py
@@ -34,9 +34,9 @@ class DownloadData:
 
         # download
         download_command = [
-            "wget",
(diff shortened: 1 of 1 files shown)


liruilong940607 (approved):
Thanks Christian! LGTM!

PR #398, opened 2024-09-10
Implement Fisheye-GS.

jefequien:
This PR follows the formulation from [Fisheye-GS](https://github.com/zmliao/Fisheye-GS) and implements new fisheye_proj and fisheye_proj_vjp functions. This allows gsplat to train with a wider FOV and thus achieve higher metrics on the ZipNerf dataset. The images in the ZipNerf dataset need to be preprocessed by remapping them to an ideal fisheye (equidistant) camera. This PR also implements masking during training.

berlin

# Results

Alameda scene with bilateral grid, fisheye, mcmc w/ 2M (CC PSNR: 23.77)

https://github.com/user-attachments/assets/4eef06cd-5191-4bb5-8f3b-640c6775bbbb


examples/simple_trainer.py
```diff
@@ -55,6 +55,8 @@ class Config:
     global_scale: float = 1.0
     # Normalize the world space
     normalize_world_space: bool = True
+    # Camera model
+    camera_model: str = "pinhole"
```
brentyi, line 59: minor note: it would be nice to use `Literal["fisheye", "pinhole"]` here and elsewhere!

ichsan2895:
Hello, thanks for this PR. Please let me ask some question:
1. Which one dataset that get benefit of this PR? `ZipNerf fisheye` or `ZipeNerf Undistorted`?
image

2. If I convert `ZipNerf fisheye` dataset using `colmap image_undistorter`, then I train it using `--camera_model fisheye`. Do it get boost the quality?

examples/datasets/colmap.py
```diff
@@ -236,16 +237,62 @@ def __init__(
...
+                            1.0
+                            + params[0] * theta**2
+                            + params[1] * theta**4
+                            + params[2] * theta**6
+                            + params[3] * theta**8
+                        )
+                        x2 = fx * x1 * r + width // 2
+                        y2 = fy * y1 * r + height // 2
+                        mapx[j, i] = x2
+                        mapy[j, i] = y2
```
liruilong940607, line 273: batchify this?

jefequien: Done

jefequien:
@ichsan2895 Use the dataset download script to download the fisheye dataset and then run the benchmarking script.
```
python datasets/download_dataset.py --dataset zipnerf
bash benchmarks/fisheye/mcmc_zipnerf.sh
```

liruilong940607 (approved):
This is super cool! Thanks for consolidating the camera_model. LGTM!

gradeeterna:
Amazing work, can't wait to try!

How can we remap our own fisheye images to an ideal fisheye equidistant camera?

I use a Samyang 8mm lens on APS-C, which is very similar to the ZipNeRF dataset (12mm fisheye on full-frame I think) and also circular fisheyes from 360 cameras, which need the black edges masked out.

Thanks!

jefequien:
The dataloader will do it automatically. You just need to use COLMAP to provide the fisheye distortion parameters.
https://colmap.github.io/cameras.html

Support for masks right now is not great. You will need to follow the code and feed your own masks into the training process.

scott198510:
@jefequien  I follow your comments, 
python datasets/download_dataset.py --dataset zipnerf
bash benchmarks/fisheye/mcmc_zipnerf.sh
The video background after my training and rendering is very dim. 
What's going on? Obviously not as bright as the video you provided above, but the rendering result of the image looks normal
alameda_ellipse
nyc_ellipse
berlin_ellipse
london_ellipse
All rendering  results look very dim,

szhang963:
Hi, I have a question about fisheye render.
I got an imcomplete render result when I used a larger fov, like this,
image
The left is original intrinsic, and the right is new intrinsic of value,
`
K = [[439.2054,   0.0000, 961.7526],
         [  0.0000, 439.5367, 762.4213],
         [  0.0000,   0.0000,   1.0000]]
width = 1920
height = 1536
`
Could you tell me how to render a complete result that not have black edge?
Thanks in advance.

szhang963:
@liruilong940607 hi, could you help me take a look at this question？

liruilong940607:
@szhang963 sorry just saw this. It feels normal to me to have black surroundings. That is just how fisheye barrel distortion looks like. Illustration on this page might be helpful:
https://commonlands.com/blogs/technical/wide-angle-lenses-and-fisheye-camera-lens-distortion

liruilong940607:
@szhang963 I probably don't get why would you like to render with negative z, and unfortunately I probably don't have capacity recently too look into this.

But a quick thought is that we assume `z` is always positive here
https://github.com/nerfstudio-project/gsplat/blob/d09ba43aa98e69023d4104efc318bf4b4b04b8cf/gsplat/cuda/include/Utils.cuh#L572

If consider z might be negative (tho i don't know why that is a valid case), we should update that line of code to
```
 float theta = glm::atan(xy_len, std::abs(z) + eps); 
```

Maybe you could try this and let me know if it resolve your issue?

liruilong940607:
what would it look like if you render with pinhole?

szhang963:
I only modifies _fully_fused_projection from _torch_impl. Could you give me some guidances for debugging rasterization function in order to check render error when z-near < 0?

szhang963:
I render `pinhole `camera in z-near = 0.01 and -1, the results are below,
image
Therefore, I think it is not reasonable for rendering a large of black area though z near < 0.

szhang963:
Thanks. I will try it.

PR #420, opened 2024-09-25
Make commands easier in README

MrNeRF:
Not sure if this is to your liking, but I prefer handling everything from the root. You can simply paste these commands into Bash and run them. It streamlines the process.

If this approach doesn't align with your preferences, feel free to disregard this PR.


liruilong940607 (approved):
This actually makes more sense to me as well. Thanks @MrNeRF ! Merging it.

PR #429, opened 2024-09-29
A set of comments and annotations to 2DGS CUDA kernels

DylanWaken:
I have added some comments in the CUDA kernels of 2DGS to explain the math and tricks used.
This will make the code more readable for future researchers.


liruilong940607 (approved):
LGTM! Thanks for adding the comments!

Issue #425, opened 2024-09-28
[2DGS,Bug] Median depth and normals broken for batched cameras


keunhong:
Hi Team,

The median depth seems to be corrupt for the renderings beyond the first camera when using batched cameras.

Here's a small repro:

```python
import gsplat
import torch
from torch.nn import functional as F
import mediapy as media

num_gaussians = 1000
means = F.normalize(torch.randn(num_gaussians, 3, device="cuda"), dim=-1)
means[..., 2] = torch.rand(num_gaussians) * 3 + 2
scales = torch.rand_like(means) * 0.1 + 0.05
quaternions = torch.tensor([1.0, 0.0, 0.0, 0.0], device="cuda")[None, :].expand(
    num_gaussians, 4
)
opacities = torch.rand_like(scales[..., 0])
rgbs = torch.rand_like(means)

num_cameras = 4
viewmats = torch.eye(4, device="cuda")[None].expand(num_cameras, 4, 4)
Ks = torch.tensor(
    [
        [300.0, 0.0, 150.0],
        [0.0, 300.0, 100.0],
        [0.0, 0.0, 1.0],
    ],
    device="cuda",
)[None, :, :].expand(num_cameras, 3, 3)
width, height = 300, 200

features, alphas, normals, surf_normals, dist_loss, depth_median, _ = gsplat.rasterization_2dgs(
    means=means,
    quats=quaternions,
    opacities=opacities,
    colors=rgbs,
    scales=scales,
    viewmats=viewmats,
    Ks=Ks,
    width=width,
    height=height,
    render_mode="RGB+D",
)
media.show_images(features[..., :3].numpy(force=True))
media.show_images(features[..., 3:].numpy(force=True), vmin=0, vmax=5)
media.show_images(depth_median.numpy(force=True), vmin=0, vmax=5)
media.show_images(normals.numpy(force=True)*0.5+0.5)
media.show_images(surf_normals.numpy(force=True)*0.5+0.5)
```

![image](https://github.com/user-attachments/assets/8da4f0cc-6d2b-40b5-8b46-b5ad4bfd2e56)

FantasticOven2:
Thanks for pointing this out! I will take a look.

liruilong940607:
Should be fixed in #430

keunhong:
@liruilong940607 any chance you could do a minor version release with the fix?

PR #430, opened 2024-09-30
Fix 2DGS for Correct Normal and Median Depth Calculation

khokao:
Fixes the calculation of normal and median depths in 2DGS.

issue: https://github.com/nerfstudio-project/gsplat/issues/425


khokao:
@liruilong940607  
Thanks for the review! 

> Would you mind run the formatter to pass the test?

I have formatted the code 🙏 
https://github.com/nerfstudio-project/gsplat/pull/430/commits/f839a41699aaf593dc61ef7543ccd44351f5f886

PR #431, opened 2024-09-30
densify after optimizer step. fixes #426

### LINKED ISSUE #426: [Bug] Sparse Tensors leading to incorrect gradients post densification.
I think that when `packed=True` is passed to the trainer, the gradient is wrong immediately after densification. The densification duplicates gaussians and prunes them leading to a change in the order and the count of the gaussians.

In [this line](https://github.com/nerfstudio-project/gsplat/blob/05829ef9ce4780a82e75314950f567be7ef857ad/examples/simple_trainer.py#L756), a sparse tensor is created with the densified gaussians. But the `gaussian_ids` have been calculated according to the pre-densification configuration.

This can also be verified by simply printing the maximum value of the `gaussian_ids` and the shape of the gaussian tensors post-densification. I can see that the maximum value of `gaussian_ids` is greater than the shape of the post-densification tensors.

```
(Pdb) print(self.splats[k].size())
torch.Size([52346, 3])
(Pdb) print(gaussian_ids.max())
tensor(54274, device='cuda:0')
```

Can someone also verify my thought process?

rahul-goel:
Fixes #426.

I don't think it will have a change in the final quality metrics at all. Haven't tested though.

### DIFF
--- a/examples/simple_trainer.py
+++ b/examples/simple_trainer.py
@@ -724,27 +724,6 @@ class Runner:
                     data, f"{self.ckpt_dir}/ckpt_{step}_rank{self.world_rank}.pt"
                 )
 
-            if isinstance(self.cfg.strategy, DefaultStrategy):
-                self.cfg.strategy.step_post_backward(
-                    params=self.splats,
-                    optimizers=self.optimizers,
-                    state=self.strategy_state,
-                    step=step,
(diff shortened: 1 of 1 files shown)


liruilong940607 (approved):
Thanks this LGTM!

PR #432, opened 2024-09-30
Selective Adam

rahul-goel:
This is the same as the sparse adam described in the Taming3DGS paper. Right now the non-zero radii gaussians are optimized. Hence the flag `visible adam` is used in the command line argument. Any other kind of mask can be passed to the same optimizer backend for different behaviour.

This optimizer is compatible with `packed=True` as well. It gives a speed-up as briefly evaluated in the table below. I tried the bicycle scene till 7000 iterations on 3080Ti. Although, it leads to a different number of gaussians and slightly different evaluation metrics. The speed-up should be higher and more noticeable with higher number of iterations (due to higher number of gaussians).

|                     | Adam    | PyTorch Sparse Adam + Packed | Visible (Selective) Adam | Visible (Selective) Adam + Packed |
|---------------------|---------|------------------------------|--------------------------|-----------------------------------|
| Time                | 3m 16s  | 3m 16s                       | 2m 28s                   | 2m 37s                            |
| Count | 2642634 | 2234936                      | 2325614                  | 2337493                           |

PS: I've created a new optimizer directory similar to strategy directory. I expect different optimizers catering to 3DGS to show up as research progresses. So I think this is a good option.

### DIFF
--- a/examples/simple_trainer.py
+++ b/examples/simple_trainer.py
@@ -764,7 +764,7 @@ class Runner:
             # optimize
             for optimizer in self.optimizers.values():
                 if cfg.visible_adam:
-                    optimizer.step(visibility_mask, gaussian_cnt)
+                    optimizer.step(visibility_mask)
                 else:
                     optimizer.step()
                 optimizer.zero_grad(set_to_none=True)
--- a/examples/simple_trainer.py
+++ b/examples/simple_trainer.py
@@ -40,7 +40,7 @@ from gsplat.compression import PngCompression
 from gsplat.distributed import cli
 from gsplat.rendering import rasterization
 from gsplat.strategy import DefaultStrategy, MCMCStrategy
-from gsplat.cuda._wrapper import SelectiveAdam
+from gsplat.optimizers import SelectiveAdam
 
 
 @dataclass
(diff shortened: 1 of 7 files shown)


rahul-goel:
Help from the community to evaluate the performance and final evaluation metrics for 30K iterations and to test this for multi-GPUs will be very much appreciated. :)

rahul-goel:
I was able to test the `garden` scene on a bigger GPU. Here are the results with the default strategy:

|       | Adam    | Visible (Selective) Adam |
|-------|---------|--------------------------|
| Time  | 1821s   | 1452s                    |
| Count | 5771257 | 4869387                  |
| PSNR  | 27.343  | 27.259                   |
| SSIM  | 0.866   | 0.862                    |
| LPIPS | 0.076   | 0.082                    |

The comparison can be unfair here due to different number of gaussians. So I tried the MCMC strategy with 1M gaussian cap and the results for that are:

|       | Adam  | Visible (Selective) Adam |
|-------|-------|--------------------------|
| Time  | 1272s | 1188s                    |
| Count | 1M    | 1M                       |
| PSNR  | 26.97 | 26.78                    |
| SSIM  | 0.848 | 0.839                    |
| LPIPS | 0.113 | 0.122                    |

But here, I think that the densification calculation is slow which hides the speed-up of selective adam.

Finally, we can see a drop in metrics. I think the main reason for this is that this is a different optimizer and requires different learning rates which I'm avoiding due to extra complexity. From my old experiments on the outdoor scenes of MN360, changing the learning rates led to better quality.

liruilong940607:
This is cool! Do you want to add a doc string to the `SelectiveAdam` and maybe cite the Taming GS paper there?

rahul-goel:
Done!

liruilong940607 (approved):
LGTM now!

PR #445, opened 2024-10-09
Correcting 3 sigma in 2D gaussian distribution

jckhng:
2DGaussian
2D multivariate distribution has a different probability related to 1-2-3 sigma 

The black scatter is a plot of 1000 samples 2D multivariate gaussian distribution. The blue ellipse uses the 3 sigma scaling. 3 sigma scaling works well in 1 dimension, and correspond to 99.7% probability. However, the corresponding probability in 2D is actually lower if 3 sigma scaling is used. To get the equivalent 99.7% red ellipse in 2D, it should be scaled by 3.4086 instead. 

This is also described in the table in this [link](https://en.wikipedia.org/wiki/Multivariate_normal_distribution#Geometric_interpretation)

Thus, I propose to change the scale_factor to 3.4086, to have the same effect of the 99.7% ellipse such that the contribution of the gaussian beyond the ellipse boundary becomes vanishingly small.

Also modified the inside view frustum test to use the AABB of the ellipse directly.


jb-ye:
Did you observe any material improvements in terms of quality?

PR #504, opened 2024-11-25
Fix proj.cuh include guard

desmondlzy:
The include guard's name in `proj.cuh` appears to be inconsistent with the file's. Might cause minor building issues when trying to include this header.


liruilong940607 (approved):
Thanks for the fix! Sorry for the late response!

PR #508, opened 2024-11-28
fix: wrong torch cat dim when concatenating colors+depths in simple_trainer_2dgs.py

Crescent-Saturn:
A minor type in `example/simple_trainer_2dgs.py`

It was a useless `if` condition there. I thought you wanted to concatenate `colors` and `depths` along different directions based on their shape for better visualization:
- Vertical if `width > height`
- Horizontal else

Thanks for your time!


liruilong940607 (approved):
Nice Catch!

Issue #506, opened 2024-11-26
Batch evaluation script doesn't work out of the box


jerinphilip:
Maybe I'm missing something, but I had to do the following to get the command from README.md running:

```bash
# run batch evaluation
bash examples/benchmarks/basic.sh
```

```diff
diff --git a/examples/benchmarks/basic.sh b/examples/benchmarks/basic.sh
index 6b043c56..cd53d1f2 100644
--- a/examples/benchmarks/basic.sh
+++ b/examples/benchmarks/basic.sh
@@ -1,6 +1,7 @@
 SCENE_DIR="data/360_v2"
 RESULT_DIR="results/benchmark"
 SCENE_LIST="garden bicycle stump bonsai counter kitchen room" # treehill flowers
+EXAMPLES_DIR="examples"
 RENDER_TRAJ_PATH="ellipse"
 
 for SCENE in $SCENE_LIST;
@@ -14,7 +15,7 @@ do
     echo "Running $SCENE"
 
     # train without eval
-    CUDA_VISIBLE_DEVICES=0 python simple_trainer.py default --eval_steps -1 --disable_viewer --data_factor $DATA_FACTOR \
+    CUDA_VISIBLE_DEVICES=0 python ${EXAMPLES_DIR}/simple_trainer.py default --eval_steps -1 --disable_viewer --data_factor $DATA_FACTOR \
         --render_traj_path $RENDER_TRAJ_PATH \
         --data_dir data/360_v2/$SCENE/ \
         --result_dir $RESULT_DIR/$SCENE/
@@ -22,7 +23,7 @@ do
     # run eval and render
     for CKPT in $RESULT_DIR/$SCENE/ckpts/*;
     do
-        CUDA_VISIBLE_DEVICES=0 python simple_trainer.py default --disable_viewer --data_factor $DATA_FACTOR \
+        CUDA_VISIBLE_DEVICES=0 python ${EXAMPLES_DIR}/simple_trainer.py default --disable_viewer --data_factor $DATA_FACTOR \
             --render_traj_path $RENDER_TRAJ_PATH \
             --data_dir data/360_v2/$SCENE/ \
             --result_dir $RESULT_DIR/$SCENE/ \

```

PR #510, opened 2024-12-03
Fix evaluation scripts

rjbaw:
This should fix evaluation paths without modifying the script itself nerfstudio-project/gsplat#506  and also fix 
"from gsplat.optimizers import SelectiveAdam" not found


batchor:
Same with #509 .

@liruilong940607 @vye16 This is an obvious error and could be easily fixed.

maturk (approved):
LGTM

PR #541, opened 2025-01-17
bugfix in transmittance based early stop criteria

wonjune23:
I think there's a bug in terms of when to stop rendering if the transmittance gets near zero.
Intuitively, the stopping criteria checks if the transmittance for rendering "next Gaussian" is very low or not.
Therefore, it should render the "current Gaussian" and then break out of the loop.
However, the current implementation instead checks the criteria and stop immediately, without rendering the current one.

To be more specific, say there are only two Gaussians to be rendered for a pixel, where the first one to be rendered has alpha of 0.5, and the other one has alpha of 1.0. Obviously, because the second Gaussian has alpha of 1.0, the rendered alpha of the pixel should be 1.0.
However, the current implementation only renders the first one because while trying to render the second one, it detects that T will get near zero and doesn't render the second one, leaving the pixel's rendered alpha 0.5.

This can be verified by manually rendering two very bright Gaussians (with the second one being very large):

```
self.means = torch.tensor([[0, 0, 1],
                            [0, 0, 1.5]], device=self.device)
self.scales = torch.tensor([[1.5e1, 1.5e1, 1.5e1],
                            [1e9, 1e9, 1e9]], device=self.device)
d = 3
self.rgbs = torch.ones(self.num_points, d, device=self.device) * 1e9

u = torch.ones(self.num_points, 1, device=self.device)
v = torch.ones(self.num_points, 1, device=self.device)
w = torch.ones(self.num_points, 1, device=self.device)

self.quats = torch.cat(
    [
        torch.sqrt(1.0 - u) * torch.sin(2.0 * math.pi * v),
        torch.sqrt(1.0 - u) * torch.cos(2.0 * math.pi * v),
        torch.sqrt(u) * torch.sin(2.0 * math.pi * w),
        torch.sqrt(u) * torch.cos(2.0 * math.pi * w),
    ],
    -1,
)

self.opacities = torch.tensor([1e9, 1e9], device=self.device)
```

It renders this weird gray ball, while after applying this bugfix, it correctly renders white image.
|  current | bugfix |
|:--:|:--:|
| current | bugfix |

If you cut and look at the middle section of the above image, it looks like this:

|  current | bugfix |
|:--:|:--:|
| current_midsection  |  bugfix_midsection |

Here I made randomly initialized Gaussians to learn an all-white image. This behavior compromises stability of training, and it makes the 3DGS never be able to learn a truly saturated image. You can see this from the loss not being able to get reduced below a certain point. The output image has slightly different color because the current implementation struggles to render a truly white image.
|  current | bugfix |
|:--:|:--:|
| current_training  | training |
| current_loss | bugfix_loss |

If you discard the color values below 250 and map [250, 255] to be [0, 255] to see the effect better, you can clearly see the current implementation struggles because of the weird gray blobs. 

|  current | bugfix |
|:--:|:--:|
| current_training_enhanced |training_enhanced |
- (please ignore the RGB being crazy at the beginning. I didn't really discard the values below 250 so they became minus and got repeatedly underflown.)

Though I think this bug has minimal effect in practice, I still think it needs a fix ;)

### DIFF
--- a/gsplat/cuda/csrc/rasterize_to_pixels_2dgs_fwd.cu
+++ b/gsplat/cuda/csrc/rasterize_to_pixels_2dgs_fwd.cu
@@ -302,12 +302,6 @@ __global__ void rasterize_to_pixels_fwd_2dgs_kernel(
                 continue;
             }
 
-            const S next_T = T * (1.0f - alpha);
-            if (next_T <= 1e-4) { // this pixel is done: exclusive
-                done = true;
-                break;
-            }
-
             // run volumetric rendering..
             int32_t g = id_batch[t];
             const S vis = alpha * T;
@@ -343,6 +337,13 @@ __global__ void rasterize_to_pixels_fwd_2dgs_kernel(
             }
 
             cur_idx = batch_start + t;
+
+            const S next_T = T * (1.0f - alpha);
+            if (next_T <= 1e-4) { // this pixel is done: exclusive
+                T = next_T;
+                done = true;
+                break;
+            }
 
             T = next_T;
         }
--- a/gsplat/cuda/csrc/rasterize_to_pixels_fwd.cu
+++ b/gsplat/cuda/csrc/rasterize_to_pixels_fwd.cu
@@ -147,12 +147,6 @@ __global__ void rasterize_to_pixels_fwd_kernel(
                 continue;
             }
 
-            const S next_T = T * (1.0f - alpha);
-            if (next_T <= 1e-4) { // this pixel is done: exclusive
-                done = true;
-                break;
-            }
-
             int32_t g = id_batch[t];
             const S vis = alpha * T;
             const S *c_ptr = colors + g * COLOR_DIM;
@@ -161,6 +155,13 @@ __global__ void rasterize_to_pixels_fwd_kernel(
                 pix_out[k] += c_ptr[k] * vis;
             }
             cur_idx = batch_start + t;
+
+            const S next_T = T * (1.0f - alpha);
+            if (next_T <= 1e-4) { // this pixel is done: exclusive
+                T = next_T;
+                done = true;
+                break;
+            }
 
             T = next_T;
         }


yzslab:
There is a bug in your modification. The `T = next_T` should be placed before the `break`, otherwise, the `render_alphas` and `render_colors` will use the wrong transmittance.

lopeLH:
Note that the behavior you are trying to fix is not a bug, but an intentional feature introduced by the original 3DGS authors to avoid numerical stability issues. With your changes, the final transmittance can be arbitrarily close to zero, causing numerical stability issues in the backward pass.

From annex C of the original 3DGS paper:

image

wonjune23:
@lopeLH Hi, thank you for letting me know, I was not aware if it was intended!

However, I've thought about all this and it looks like the only concern in terms of numerical stability with `T` is that we need to restore it starting from `T_final` by repeatedly dividing it by `1-alpha`:
```cpp
S ra = 1.0f / (1.0f - alpha);
T *= ra;
```

 I argue that it's not really a problem in current implementation because of these reasons:

1. maximum of `alpha` is truncated to be `0.999` : 
```cpp
S alpha = min(0.999f, opac * __expf(-sigma));
```
2. minimum of `alpha` is `1/255` (otherwise, the renderer just ignores it): 
```cpp
if (sigma < 0.f || alpha < 1.f / 255.f) {
    continue;
}
```
3. The minimum possible value of `T` will be `1e-7` after my bugfix (before the bugfix, it was `1e-4`): 
    - This will happen when there are several different Gaussians rendered so that `T` became near `1e-4`,  and after that when there is the last Gaussian that would not have been rendered before the bugfix, with its `alpha = 0.999`. This is the case where `T` becomes near `1e-7`. The point is that the `T` still has bounded minimum value. 

4. Numerical error of restoring `T` is also well bounded and is negligible.
    - There will be the largest numerical drift when we iteratively divide the `T` with near `1.0` values (i.e., small `alpha`), so that the backward pass will perform the division as many times as possible. This maximum number of division operation is also bounded, because `alpha` is bounded by `1/255`.
    - Now, I've tried some experiment, and the current implementation can handle the worst situation with negligible numerical errors. Specifically, in the backward pass when we restore the `T` from `T_final`, say we start from `T_final = 1e-7`, the wors case scenario. This implies that the last Gaussian's `alpha` was actually `0.999`, so we divide `T` with `0.001` and get `T=1e-4`. 
    - From here on it's the same situation before the bugfix was applied, but I still tried the worst case scenario: all the Gaussians had `alpha=1/255`. In this case, there can be way less than 2400 Gaussians because if we divide the `T=1e-4` with `1/(1-1/255)` for 2400 times, it gets larger than `1.0`.

Specifically, the experiment was like this:
```cpp
S temp_T = (1e-4) * (1.0 - 0.999);
temp_T /= (1.0 - 0.999);
for (int l = 0; ; < 2400; l++) {
    temp_T /= (1.0 - 1.f / 255.f);
}
printf("%f\n", temp_T);
```

In the current implementation `S` is single precision float. This prints `1.245986`. 
If we perform this in Python (that uses double precision float), it prints `1.2459859743472113`.

Therefore, my conclusion is that even in the worst case scenario, my bugfix doesn't affect the numerical stability regarding `T`.

I feel like this PR is getting an overkill for such a negligible bug, but again, I still think it's a bug. If something is wrong with my logic, please let me know :)

lopeLH:
I am still not fully convinced about the proposed changes, for a number of reasons:

-  From a software engineering perspective, I find it dangerous that the change's numerical stability relies on the values of other hyper-parameters which might be changed in the future, inadvertently triggering numerical stability issues. In this regard, I think a safer solution could involve providing the backward pass with the penultimate transmittance, instead of the last one, to bypass the issue altogether.

- From an algorithmic point of view, I think the change is less trivial and safe than it seems, and might affect the training dynamics in unforeseen ways. Adopting it safely would require careful validation and QA, as it is very easy to miss small but critical details in such a complex codebase. For example, I think your numerical stability analysis overlooked the fact that transmittance is converted to opacity (`1.0f. - T`) and back to transmittance when passed from forward to backward kernels. I think you will find the numerical error is not negligible if you consider that in your analysis.

- Don't know what the purpose of this library is. If the goal is to reproduce the original implementation results as faithfully as possible, even if the other concerns are addressed one could argue this change deviates from the original author's proposal. This is more of an "improvement" proposal than a trivial bugfix.

That said, I am not a maintainer of this library, so these are just my 2cents.

wonjune23:
Hi, I agree on all three points, especially the second point: as you mentioned, I overlooked the `1.0f -T` part, and it indeed makes the error not negligible.

I also felt like this is getting deviated too much from the original implementation. As I mentioned earlier, I was not aware if it was intended in the original paper. After knowing that, I agree that even withought the concerns it's going too far from the original one. But after that, I still thought this feature, although intended, put too much constriant on the rendering process than what's required. But as you pointed out, it was not too much.

I still am not happy with the fact that in some cases, despite being very rare, the rendered image can actually be very different than what's expected. However, it turns out the fix for this is not that easy without further changing the current method. But thank you for sharing all these. Knowing this might actually be very helpful on what I was working on, and hopefully to others as well.

PR #543, opened 2025-01-20
minor fix to 2dgs corner case

FantasticOven2:


FantasticOven2:
> What does this fix?

Hi, I changed some tensor initializations to avoid `nan` access

gsplat/rendering.py
```diff
@@ -1233,7 +1233,9 @@ def rasterization_2dgs(
     # Rasterize to pixels
     if render_mode in ["RGB+D", "RGB+ED"]:
         colors = torch.cat((colors, depths[..., None]), dim=-1)
-        # backgrounds = torch.cat((backgrounds, torch.zeros((C, 1), device="cuda")), dim=-1)
+        backgrounds = torch.cat(
+            (backgrounds, torch.ones((C, 1), device=colors.device)), dim=-1
+        )
     elif render_mode in ["D", "ED"]:
```
lionlai1989, line 1239: Hi, why is `zeros` replaced with `ones`? I think `zeros` is also initialized?

brentyi (approved):
makes sense to me!

gsplat/rendering.py
whikwon: When running `bash ./benchmarks/basic_2dgs.sh`, I encounter an error indicating that `backgrounds` is set to None.

PR #559, opened 2025-02-13
Fix channel padding for distortion and median depth computation

jonathsch:
The CUDA kernel requires the depth information used for calculating distortion loss and median depth to be located in the last channel of the tensor. However, because of zero-padding, the depth values for each Gaussian are no longer stored in the last channel. This results in incorrect computations for both the distortion loss and the median depth.


liruilong940607 (approved):
Nice Catch!

liruilong940607 (approved):

PR #569, opened 2025-02-25
fix the offset of K in projection packed fwd 2dgs kernel

lapisliu:
Fix the offset of K in projection packed fwd 2dgs kernel by adding `Ks += row_idx * 9`.


liruilong940607 (approved):
Nice catch!

PR #583, opened 2025-03-16
refactor CUDA code for faster compilation and easier extensibility

liruilong940607:
Faster compilation: 

- remove `#include <torch/extension.h>` from .cu files (only in .cpp files now). This makes NVCC compile significantly faster.
- Minimize the includes in .cu files.
- Use precompiled header for <torch/extension.h>.

Easier extensibility:

- C++ API: now the code is cleaned up so gsplat has an API on C++ end. Just `#include <gsplat/cuda/include/Ops.h>`, all supported functions are declared in this single file.
- Added a  null operator in `cuda/csrc/Null.{h, cpp, cu}`, as a template/tutorial for adding new operators to gsplat.

PR #585, opened 2025-03-17
Fixes build issues but only works with pytorch >= 2.2

MrNeRF:
As described here: https://github.com/nerfstudio-project/gsplat/pull/583
the pr 583 introduced some build issues. It also still does not work with pytorch < 2.2

I leave it here as a first commit to get the issues fixed.


liruilong940607:
btw i removed environment.yml in a followed up PR. As it contains many unnecessary dependencies for the gsplat library itself, while contains insufficient dependencies for the code under examples/. Hope this makes sense!

MrNeRF:
Nice, thank you! I will test it later on.

joel-evercoast:
For anyone else dealing with sudden failed builds from the git repo, whether from main or from tagged releases, the fix for me was to pin `setuptools==69.5.1`.

Newer versions of setuptools caused a variety of failures, either at build time:

`TypeError: Command.__init__() got an unexpected keyword argument 'no_python_abi_suffix'`

or afterwards getting an error about a missing `gsplat_cuda.so` file in the pytorch cache.

PR #594, opened 2025-03-24
fix save ply if else

liruilong940607:


liruilong940607:
This reduces training on bonsai (1.1M GS) from 662s to 642s on RTX 6000 Ada.

PR #622, opened 2025-04-05
faster fused-ssim

MrNeRF:
I made significant optimizations to the fused-ssim implementation. The original speed-up is available in my repo: https://github.com/MrNeRF/optimized-fused-ssim

However, the optimized code has also been merged into the official repo: https://github.com/rahul-goel/fused-ssim/pull/11

I recommend updating the dependency to this commit so gsplat can benefit from the improvements!

image

### DIFF
--- a/examples/requirements.txt
+++ b/examples/requirements.txt
@@ -19,4 +19,4 @@ tensorboard
 tensorly
 pyyaml
 matplotlib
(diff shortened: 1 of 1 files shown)


liruilong940607 (approved):
OH THIS IS GREAT!!

PR #628, opened 2025-04-09
Added Gaussian Splat Exporters

bolopenguin:
# Description

I created a snippet to export the GS models to the following  formats:
- `ply`: the original ply format supported by the vast majority of engines
- `splat`: another popular format
- `ply_compressed`: a compressed ply format supported by the playcanvas engine (ex. Supersplat)

It works with the classic GS configuration: `means`, `scales`, `quats`, `opacities`, `sh0`, `shN`

# Type of change

- [ ] Bug fix (non-breaking change)
- [X] New feature 
- [ ] Code improvement (refactoring)
- [ ] Documentation update
- [ ] Test (new or updated tests)

# Additional information

I moved the `exporter` into gsplat and updated the simple trainer to use it instead of `save_ply`.
I also added a deprecation warning to `save_ply`.

It doesn't work when using the `AppearanceOptModule` in the simple trainer, but I believe that if you're using that optimizer, exporting the GS correctly isn't possible anyway.

Let me know if you find this useful or if any changes are needed before merging!

### DIFF
--- a/examples/simple_trainer.py
+++ b/examples/simple_trainer.py
@@ -31,12 +31,12 @@ from torchmetrics.image.lpip import LearnedPerceptualImagePatchSimilarity
 from typing_extensions import Literal, assert_never
 from utils import AppearanceOptModule, CameraOptModule, knn, rgb_to_sh, set_random_seed
 
+from gsplat import export_splats
 from gsplat.compression import PngCompression
 from gsplat.distributed import cli
 from gsplat.optimizers import SelectiveAdam
 from gsplat.rendering import rasterization
 from gsplat.strategy import DefaultStrategy, MCMCStrategy
-from gsplat import export_splats
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


examples/simple_trainer.py
```diff
@@ -743,19 +743,35 @@ def train(self):
...
-                    # eval at origin to bake the appeareance into the colors
-                    rgb = self.app_module(
-                        features=self.splats["features"],
-                        embed_ids=None,
-                        dirs=torch.zeros_like(self.splats["means"][None, :, :]),
-                        sh_degree=sh_degree_to_use,
-                    )
-                    rgb = rgb + self.splats["colors"]
-                    rgb = torch.sigmoid(rgb).squeeze(0)
+                    import warnings
```
liruilong940607, line 748: Actually I think saving PLY is still doable with app_opt. The code convert features into the rgb color, and all it needs is to convert rgb to sh0 then it is ready to save

examples/simple_trainer.py
```diff
@@ -743,19 +743,35 @@ def train(self):
...
+                        quats=quats,
+                        opacities=opacities,
+                        sh0=sh0,
+                        shN=shN,
+                        format="ply",
+                    )
+                    with open(
+                        f"{self.ply_dir}/point_cloud_{step}.ply", "wb"
+                    ) as binary_file:
+                        binary_file.write(splats_bytes)
```
liruilong940607, line 774: Can we fold this into the export function? We could take in an optional argument as file name, and directly dump the bytes to it. I understand sometimes you might want to return the bytes but I feel most of the time user just want to save it to a file in one go.

liruilong940607 (commented):
Awesome work! Just left a few minor comments!

examples/simple_trainer.py
bolopenguin: Are you sure about that?
Because when I am launching the benchmark on main branch with both options `save_ply` and `app_opt` I get the following error:
```
    sh0 = numpy_data["sh0"].transpose(0, 2, 1).reshape(means.shape[0], -1)
KeyError: 'sh0'
```
And it makes sense to me, if you are using app_opt you do not have the spherical harmonics in the splat dictionary

liruilong940607: Sorry the code was buggy before. 

Yes with app_opt the best we can do is to export color **without** view-dependent effect. I.e., only use `sh0` and setting `shN` to all zero or just an empty tensor with shape [N, 0, 3] (does the exporter work with [N, 0, 3] shape?)

`sh0` could be computed from `rgb` easily:
```
sh0 = (rgb - 0.5) / C0
```

gsplat/__init__.py
```diff
@@ -18,6 +18,7 @@
     world_to_cam,
 )
 from .optimizers import SelectiveAdam
+from .exporter import export_splats
```
liruilong940607, line 21: sorry forget to mention this that we also need to add it to the `all=[]` field below.

examples/simple_trainer.py
bolopenguin: Good catch! I made the changes you suggested, and it works correctly now.

PR #640, opened 2025-04-17
:bug: Fix memory leak in exporters - splat2ply bytes

mrcabellom:
Fix memory issue in ply gaussian export by replacing struct.pack with efficient tensor serialization using numpy

### DIFF
--- a/gsplat/exporter.py
+++ b/gsplat/exporter.py
@@ -3,6 +3,7 @@ import struct
 from io import BytesIO
 from typing import Literal, Optional
 
+import numpy as np
 import torch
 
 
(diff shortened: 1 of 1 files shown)


bolopenguin (commented):
I did not use numpy because the endianess [depends on the machine architecture](https://numpy.org/devdocs/user/byteswapping.html?utm_source=chatgpt.com).

So pay attention with that, even if the most common architecture is big endian.
You should use `dytpe='>f4'` instead, but I am not sure about it.

And if it works, it would be nice to have it too for the other two formats: `ply_compressed` and `splat`

mrcabellom:
Thanks for the comment! Just to clarify:

You're right that endianness can be important when writing binary files, but in this case, most modern architectures (x86, x64) are actually little-endian, not big-endian. When no byte order is specified in NumPy, it defaults to the native endianness of the system 

I've now made the byte order explicitly little-endian in all relevant outputs to avoid ambiguity and ensure cross-platform consistency.

In addition, I've optimized the splat2ply_compressed and splat2splat functions to eliminate memory leaks and avoid inefficient Python object conversions. The resulting files have been tested in [SuperSplat Editor](https://superspl.at/editor) and load correctly without issues.

liruilong940607:
This seems nice. I'm not familiar with endianness stuffs but seems the latest commit is good? @bolopenguin Would you mind help reviewing again? 

@mrcabellom Also could you run the formatter so that tests could pass?
https://github.com/nerfstudio-project/gsplat/blob/86dd00c69b71d956d70f581e08b96829b22b48ea/formatter.sh#L7-L8

mrcabellom:
@liruilong940607 I just updated the file using the Black formatter. Sorry about that, I hadn’t looked at the contribution guidelines in the docs/Dev.md beforehand.

bolopenguin (approved):
LGTM
Great work, now it is much faster!

liruilong940607 (approved):

liruilong940607:
Thank you both for the awesome work!

PR #642, opened 2025-04-18
Fixed redundant PyTest mark that was applied to a fixture.

findeevy:
Fixed redundant mark that was applied to a fixture that had no effect and was throwing a warning from PyTest (CUDA/Pytorch installs worked).

Screenshot from 2025-04-18 11-37-05


liruilong940607 (approved):
LGTM!

PR #651, opened 2025-04-24
Fix _jit_compile arguments for torch>=2.7.0

shiukaheng:
PyTorch 2.7.0 changed the _jit_compile signature to support SYCL, compiling as is now prevents gsplat from JIT compilation. Added a fallback case to calling _jit_compile to handle this while being backwards compatible.


liruilong940607:
Thanks @shiukaheng this looks great! Would you mind running the formatter so it pass the tests?

miriameng:
I can confirm this fixes torch 2.7.0, can we merge this?

liruilong940607:
LGTM! Thanks for the fix!

Issue #660, opened 2025-05-01
[BUG] Empty ".ply" file upon encountering NaN in splats["opacities"]


Crezle:
The "invalid_mask" is currently too strict if splats["opacities"] has shape (N,), as using "np.isnan(opacities).any(axis=0)" causes one NaN to filter away all the data.

https://github.com/nerfstudio-project/gsplat/blob/85d4d7aea56c0a44be4d1f164186e5f98f7b2116/gsplat/utils.py#L38-L39

Could someone verify if this is actually a bug or if I am assuming a wrong shape for the input "splats["opacitites"]"?

Thanks in advance!

liruilong940607:
Yes you are correct it's a bug! `any(axis=0)` should go away for opacities. do you want to PR it?

Crezle:
Created small PR at #663 👍

PR #663, opened 2025-05-03
[BUG] Fix strict "invalid mask" in gsplat.utils.save_ply(...)

Crezle:
**Problem**:
Current mask will remove all data if any opacities are NaN.
I assume this was applied due to similarity to the other "splats" properties that are multidimensional,
despite opacities being 1-dimensional.

**Fix**:
Remove "any()" from opacities.


Crezle:
Related to issue #660

liruilong940607 (approved):
LGTM!

PR #669, opened 2025-05-07
Implementation 3D Convex Splatting: Radiance Field Rendering with 3D Smooth Convexes (3DCS)

AlexMorgand:
This is the reimplementation of [3D Convex Splatting: Radiance Field Rendering with 3D Smooth Convexes](https://convexsplatting.github.io/)

Here are the main contributions:

- Adding convex points, delta, sigma
- Dedicated projection, rasterization and spherical harmonics computation
- Dedicated densification (only split and prune) strategy
- Saving as pt instead of ply. Output from official implementation can be loaded as well (WIP for the splat export of the new viewer)
- Adding light and outdoor mode and adapt the hyperparameters depending on the situation

How to test:

Here is an example for MipNerf360/garden

```
CUDA_VISIBLE_DEVICES=0   python examples/simple_trainer_3dcs.py 3dcs  --data-dir  path/to/MipNerf/360_v2/kitchen --result-dir /tmp/garden --max-steps 30000 --data_factor 4
```

**[WIP] Metrics are quite similar to the paper but since I had to rebase on top of the new version, I need to recompute all of them.**

Garden: PSNR: 26.611, SSIM: 0.8459, LPIPS: 0.097 Time: 0.063s/image Number of GS: 4473125

This work was approved and supervised by two authors of the paper @rvandeghen and @heldJan

image
image

### DIFF
--- a/examples/simple_viewer.py
+++ b/examples/simple_viewer.py
@@ -12,7 +12,7 @@ import viser
 from pathlib import Path
 from gsplat._helper import load_test_data
 from gsplat.distributed import cli
-from gsplat.rendering import rasterization
+from gsplat.rendering import rasterization, rasterization_3dcs
 
 from nerfview import CameraState, RenderTabState, apply_float_colormap
 from gsplat_viewer import GsplatViewer, GsplatRenderTabState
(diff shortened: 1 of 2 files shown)


liruilong940607:
Hi @AlexMorgand thanks for the effort on implementing it in gsplat!

I'm not familiar with this paper -- by skimming through it I don't get what benefit we got from this convex formulation. It is certainly a more flexible representation than GS but it seems to train & render much slower with only marginal quality improvement. 

Would you mind first elaborate the motivation of bringing it into gsplat?

imlixinyang:
@liruilong940607 Hi Ruilong, from the perspective of my personal research, I would like to explain the necessity of this method which has more parameters for primitives, degraded rendering speed, but a smaller number of primitives. I use gsplat in my research for feed-forward formed 3DGS. Due to the large number of viewpoints and pixels, the number of gs is also large. However, if we can use a smaller number but slightly more complex forms of primitives, the method will be more efficient. Because the time consumption of rendering during training is only a small part compared to the time consumption of network feed-forwarding. I believe that the introduction of this type of method will significantly benefit gsplat in more related fields beyond 3D reconstruction based on optimization.

@AlexMorgand, thank you very much for your contribution. I'm really looking forward to the day when this pull request can be merged.

imlixinyang:
Hey @liruilong940607, I wonder if your recent views on integrating such enhanced primitive representation capabilities into gsplat have changed.

You can see that more and more feed-forward reconstruction models are using gsplat for its convenient functions and support for multi-GPU distributed computing. Such work may be limited by the number of gs in the future, so they will instead rely on improving primitive representation capabilities.

For example, recent work: https://openreview.net/forum?id=RExGGPD5tw

PR #672, opened 2025-05-07
Increase maximum allowed far plane in viewer

vincentwoo:
In large scenes, 100 is a good default but can sometimes be a bit too small.


liruilong940607:
LGTM!

PR #682, opened 2025-05-11
HoGS: Unified Near and Far Object Reconstruction via Homogeneous Gaussian Splatting

MrNeRF:
This PR implements [this paper](https://arxiv.org/abs/2503.19232), which improves background reconstruction while preserving foreground fidelity compared to training without homogeneous coordinates.

I’d appreciate community support to verify this, but based on my tests, the results should be identical to the reference implementation. Currently, only the default strategy is supported. I haven't run the benchmarks yet, so I open this pr as draft.

You can find explanations and more results on their project page: https://kh129.github.io/hogs/

image

### DIFF
--- a/examples/simple_trainer.py
+++ b/examples/simple_trainer.py
@@ -54,13 +54,13 @@ class Config:
     render_traj_path: str = "interp"
 
     # Path to the Mip-NeRF 360 dataset
-    data_dir: str = "data/360_v2/garden"
+    data_dir: str = "/home/paja/data/bike_aliked"
     # Downsample factor for the dataset
-    data_factor: int = 4
+    data_factor: int = 1
     # Directory to save results
     result_dir: str = "results/garden"
     # Every N images there is a test image
-    test_every: int = 8
+    test_every: int = 300
     # Random crop size for training  (experimental)
     patch_size: Optional[int] = None
     # A global scaler that applies to the scene size related parameters
@@ -483,8 +483,8 @@ class Runner:
 
         if cfg.use_hom_coords:
             w_inv = 1.0 / torch.exp(self.splats["w"]).unsqueeze(1)
-            means *= w_inv
-            scales *= w_inv
+            means = means * w_inv
+            scales = scales * w_inv
 
         image_ids = kwargs.pop("image_ids", None)
         if self.cfg.app_opt:
@@ -1211,7 +1211,15 @@ if __name__ == "__main__":
             ),
         ),
     }
+
     cfg = tyro.extras.overridable_config_cli(configs)
+    if cfg.use_hom_coords:
+        cfg.max_steps = 50_000
+        if isinstance(cfg.strategy, DefaultStrategy):
+            cfg.strategy.refine_stop_iter = 30_000
+            cfg.strategy.refine_every = 200
+            cfg.strategy.reset_every = 6_000
+
     cfg.adjust_steps(cfg.steps_scaler)
 
     # try import extra dependencies
--- a/examples/simple_trainer.py
+++ b/examples/simple_trainer.py
@@ -54,7 +54,7 @@ class Config:
     render_traj_path: str = "interp"
 
     # Path to the Mip-NeRF 360 dataset
-    data_dir: str = "/home/paja/data/bike_aliked"
+    data_dir: str = "/home/paja/data/alex_new"
     # Downsample factor for the dataset
     data_factor: int = 1
     # Directory to save results
@@ -552,7 +552,7 @@ class Runner:
         if cfg.use_hom_coords:
             schedulers.append(
                 torch.optim.lr_scheduler.ExponentialLR(
-                    self.optimizers["w"], gamma=0.1 ** (0.5 / max_steps)
+                    self.optimizers["w"], gamma=0.1 ** (0.005 / max_steps)
                 )
             )
         if cfg.pose_opt:
@@ -1219,6 +1219,8 @@ if __name__ == "__main__":
             cfg.strategy.refine_stop_iter = 30_000
             cfg.strategy.refine_every = 200
             cfg.strategy.reset_every = 6_000
+            cfg.strategy.refine_start_iter = 1_500
+            cfg.strategy.prune_too_big = False
 
     cfg.adjust_steps(cfg.steps_scaler)
 
--- a/examples/simple_trainer.py
+++ b/examples/simple_trainer.py
@@ -37,6 +37,7 @@ from gsplat.distributed import cli
 from gsplat.optimizers import SelectiveAdam
 from gsplat.rendering import rasterization
 from gsplat.strategy import DefaultStrategy, MCMCStrategy
+from gsplat.utils import xyz_to_polar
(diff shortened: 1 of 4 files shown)


MrNeRF:
The comparison is certainly not fair (30k vs 50k iterations). However, the default won't get there even with 50k iterations.
I will hopefully find some free cycles to run the benchmarks. 

https://github.com/user-attachments/assets/b909adcb-c5d2-459c-92f6-dedc81894546

liruilong940607:
wow this is pretty cool! Yeah we should get the benchmarks (maybe at 30k since all other things are run for 30k in this repo?) so we could be confident on how much it improves and also for future reference.

This week is kinda crazy for me. happy to review the PR more closely next week!

ichsan2895:
I found a bug when saving an PLY.

Code used:
```bash
RESULT_DIR="results/benchmark_HoGS-BS1"
SCENE="garden"
RENDER_TRAJ_PATH="ellipse"
ITERS='50000'
BS='1'

CUDA_VISIBLE_DEVICES=0 python3 examples/simple_trainer.py mcmc --eval_steps $ITERS --disable_viewer --data_factor 2 \
        --render_traj_path $RENDER_TRAJ_PATH --max-steps $ITERS --batch-size $BS \
        --data_dir data/360_v2/$SCENE/ --use-hom-coords \
        --save-ply --ply-steps 30000 50000 \
        --result_dir $RESULT_DIR/$SCENE/
```
```python
Running garden
Warning: image_path not found for reconstruction
[Parser] 185 images, taken by 1 cameras.
Downscaling images by 4x from data/360_v2/garden/images to data/360_v2/garden/images_4_png.
100%|██████████| 185/185 [00:00<00:00, 125720.39it/s]
Scene scale: 1.2265263065808911
Model initialized. Number of GS: 138766
...
...
...
Step 24800: Added 0 GSs. Now having 1000000 GSs.
Step:  29999 {'mem': 1.4372062683105469, 'ellipse_time': 591.6495430469513, 'num_GS': 1000000}
  File "/workspace/GSPLAT_160a1_HoGS/gsplat-mcmc/examples/simple_trainer.py", line 1276, in <module>
    cli(main, cfg, verbose=True)
  File "/usr/local/lib/python3.10/dist-packages/gsplat/distributed.py", line 360, in cli
    return _distributed_worker(0, 1, fn=fn, args=args)
  File "/usr/local/lib/python3.10/dist-packages/gsplat/distributed.py", line 295, in _distributed_worker
    fn(local_rank, world_rank, world_size, args)
  File "/workspace/GSPLAT_160a1_HoGS/gsplat-mcmc/examples/simple_trainer.py", line 1204, in main
    runner.train()
  File "/workspace/GSPLAT_160a1_HoGS/gsplat-mcmc/examples/simple_trainer.py", line 832, in train
    means *= w_inv
RuntimeError: a leaf Variable that requires grad is being used in an in-place operation.
```

**Additional info:**
Ubuntu 22.04 LTS
Python 3.10
Torch 2.1.2+cu118
Gsplat commit [2455a5f](https://github.com/nerfstudio-project/gsplat/pull/682/commits/2455a5fd011bfc4cccf3d7453a503dfe71bc6d3b)

MrNeRF:
Thx @ichsan2895 for testing and evaluating. Maybe the degenerated rendering is caused by the bug you reported. I fixed it but lemme check if the rendering looks correct. 

Btw, yes. HoGS primary excels at background reconstruction. So you need to look at the output visually as it is not necessarily reflected by the overall psnr. Also, 1m are a bit too low in that case. Anyway, it is optional and can be activated depending on the situation.

ichsan2895:
Oh No..
image

Why it still happen? I have modified the code and rerun the expeeriment..
image

MrNeRF:
@ichsan2895 I am not 100% sure what you are doing? Is this the simple_viewer.py? Then it is loading the ckpts which are not correctly scaled.

liruilong940607:
From @ichsan2895 it seems HoGS does really win over metrics.

But I'm curious about the visuals! If this paper really shifts more GS from foreground to background and improve the quality on background that would also be pretty nice. At least it provides this flexibility.

We should really compare in a fair setting tho -- 30k v.s. 50k doesn't makes sense.  (and considering this HoGS runs slower per iteration maybe it's actually fair to compare MCMC 50k v.s. HoGS 30k if we aim for the same wall time)

ichsan2895:
> We should really compare in a fair setting tho -- 30k v.s. 50k doesn't makes sense. (and considering this HoGS runs slower per iteration maybe it's actually fair to compare MCMC 50k v.s. HoGS 30k if we aim for the same wall time)

Yes, I have ran Gsplat MCMC-1M with 50k iters. Here is the benchmark
image


> @ichsan2895 I am not 100% sure what you are doing? Is this the simple_viewer.py? Then it is loading the ckpts which are not correctly scaled.

Yeah, Thats simple_viewer.py (Right) and mkkellog gaussian viewer (left). I will try to check and recheck that I have pointed the corrrect path,

liruilong940607:
@ichsan2895 once you have figured out the viewer stuff, would you mind sharing some comments on how you feel about the visuals? esp how much the background got improved from this approach

MrNeRF:
@ichsan2895 I never use the simple_viewer so that why I was a bit puzzled. But I fixed it now.

@liruilong940607 The visual quality is not minor, it can be quite tremendous with respect to the background. It maybe shows best with the default strategy but it certainly consumes more Gaussians. I doubt that running mcmc with 1m is a good metric here.

MrNeRF:
@liruilong940607 I pretty much know what you mean.

The paper explicitly reports different values "near vs far". I experimented quite a bit with it. In my experience this method can be a bit worse on the foreground especially when there are lots of fine details. However, it pretty much always outperforms in the background 3dgs significantly. So, there is certainly a trade-off. That's also why I made it optional in the first place.

I will try to prepare some side-by-side comparisons to visualize that.

ichsan2895:
> @ichsan2895 I never use the simple_viewer so that's why I was a bit puzzled. But I fixed it now. Please pull the latest commit.

Thanks, that is fixed now
image

MrNeRF:
I don't have the complete benchmark yet, but this is default strategy 30k steps main vs this branch. I will run bother later on also under the hogs settings for 50k iterations as specified in the paper.

https://github.com/user-attachments/assets/6e172d5e-bf80-4fcc-b63a-68bfe80596a5

https://github.com/user-attachments/assets/f513ac45-053c-4a65-85e4-b99ed4b29c3d

MrNeRF:
30k steps default strategy on main
image

30k steps default strategy HoGS this branch:
image

I will also do next the benchmark for the default strategy with the proposed 50k steps for main and this branch!

1397201241:
Nice work! I noticed a discrepancy in how scales are handled between two files:

In sample_viewer.py(load ckpts):
`
scales = torch.exp(ckpt["scales"]) * w_inv
`

In sample_trainer.py (export_splats):
`
scales = torch.log(torch.exp(scales) * w_inv)
`
Why is there an extra torch.log() in export_splats?

PR #704, opened 2025-05-25
Alpha shouldn't be normalized or inversed?

RongLiu-Leo:
Screenshot 2025-05-24 164121
When I check this default alpha map, it looks quite counterintuitive. It means all my primitives are transparent.

I check the code and find that the alpha map is affected by depth normalize and inverse check boxes.
Per their hints, I think they should only affect the depth maps, not rgb and alpha.

This PR 
- Makes normalize and inverse check boxes enabled only when it's depth rendering.
- Switches Inverse to off by default, since having it on by default feels counterintuitive.


liruilong940607 (approved):
This makes sense. I'll merge it. thanks!

PR #705, opened 2025-05-25
Support view dependent & independent light decomposition

RongLiu-Leo:
https://github.com/user-attachments/assets/69f3a1c9-8ce3-4b1b-8b5c-80dc4618cc38

Besides to rendering full SH as RGB, this PR renders SH=0 component as the view independent diffuse map and SH>0 components as the view dependent specular map.

### DIFF
--- a/gsplat/rendering.py
+++ b/gsplat/rendering.py
@@ -46,7 +46,7 @@ def rasterization(
     packed: bool = True,
     tile_size: int = 16,
     backgrounds: Optional[Tensor] = None,
-    render_mode: Literal["RGB", "D", "ED", "RGB+D", "RGB+ED"] = "RGB",
+    render_mode: Literal["RGB", "D", "ED", "RGB+D", "RGB+ED", "Diffuse", "Specular"] = "RGB",
     sparse_grad: bool = False,
     absgrad: bool = False,
     rasterize_mode: Literal["classic", "antialiased"] = "classic",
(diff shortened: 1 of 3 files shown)


gsplat/rendering.py
```diff
@@ -481,6 +491,11 @@ def reshape_view(C: int, world_view: torch.Tensor, N_world: list) -> torch.Tenso
                 pass
     else:
         # Colors are SH coefficients, with shape [..., N, K, 3] or [..., C, N, K, 3]
+        colors = colors.clone()
```
liruilong940607, line 494: This `clone()` breaks the auto diff graph for training. (Also it is not ideal from the efficiency perspective)

I think in this case it's better to not touch this `rendering.py` file. We could just update the viewer file to support diffuse/specular.

liruilong940607 (commented):
Left some comments! I like the idea of supporting diffuse / specular in viewer!

gsplat/rendering.py
RongLiu-Leo: Yeah, it is a good idea

gsplat/rendering.py
```diff
@@ -481,6 +495,11 @@ def reshape_view(C: int, world_view: torch.Tensor, N_world: list) -> torch.Tenso
                 pass
     else:
         # Colors are SH coefficients, with shape [..., N, K, 3] or [..., C, N, K, 3]
+        # SH = 0 is the view-independent diffuse component, SH >= 1 are view-dependent specular components.
+        if render_mode in ("Diffuse", "Specular"):
+            colors = colors.clone()
+            sel = slice(1, None) if render_mode == "Diffuse" else slice(0, 1)
+            colors[..., sel, :] = 0.0  # Zero out the unwanted SH components.
```
liruilong940607, line 502: I still think this logic should live outside of the `rasterization` function (to viewer_render_fn). For three reasons:
- What if I want to rendering diffuse color + depth map? there is no "Diffuse+D" mode here. User would end up have to write this logic outside of `rasterization` function and use "RGB+D" mode anyway. (The major reason)
- I'm still not quite happy with the inplace operation `colors[..., sel, :] = 0.0` and `colors.clone()`. It's not very good to have inplace operations in the torch differentiable graph. It might break the auto diff graph or lead to unused parameters.

liruilong940607 (commented):
Only request is to move the diffuse/specular out of the `rasterization` function.

PR #707, opened 2025-05-26
Support 2dgs with the latest viewer

RongLiu-Leo:
https://github.com/user-attachments/assets/d99545a6-84b2-4836-9f39-39fdf8731269

- This PR adds support for running 2DGS with the latest version of nerfview, ensuring compatibility and functionality. Without this update, 2DGS cannot be executed properly. Render mode supports rgb, depth, normal, and alpha.

- Replaced the use of colormap with apply_float_colormap. While both provide similar logic, colormap triggers runtime errors in my environment. Suggest to unify the function usage and stick to apply_float_colormap.

### DIFF
--- a/examples/gsplat_viewer.py
+++ b/examples/gsplat_viewer.py
@@ -21,7 +21,7 @@ class GsplatRenderTabState(RenderTabState):
         "rgb", "depth(accumulated)", "depth(expected)", "alpha"
     ] = "rgb"
     normalize_nearfar: bool = False
-    inverse: bool = True
+    inverse: bool = False
     colormap: Literal[
         "turbo", "viridis", "magma", "inferno", "cividis", "gray"
     ] = "turbo"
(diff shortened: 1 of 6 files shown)


liruilong940607:
This looks cool! But the depth rendering is pretty weird -- doesn't look like depth map at all. Is the viewer accessing the correct depth data?

RongLiu-Leo:
https://github.com/user-attachments/assets/41205b5d-0462-4aed-bf92-913f8df1fe1f

Looks good now! Thanks for pointing it out.

Chuan-10:
Hi, thank you for the great work on gsplat!

I noticed a small bug in [examples/simple_trainer_2dgs.py](https://github.com/nerfstudio-project/gsplat/blob/65042cc501d1cdbefaf1d6f61a9a47575eec8c71/examples/simple_trainer_2dgs.py#L977-L1002), which seems to be related to the recent Arbitrary batching feature. The current implementation of the depth and normal rendering logic may cause bugs in some cases.

It could be modified as follows:

```python
if render_tab_state.render_mode == "depth":
    # normalize depth to [0, 1]
    depth = render_median[0]
    if render_tab_state.normalize_nearfar:
        near_plane = render_tab_state.near_plane
        far_plane = render_tab_state.far_plane
    else:
        near_plane = depth.min()
        far_plane = depth.max()
    depth_norm = (depth - near_plane) / (far_plane - near_plane + 1e-10)
    depth_norm = torch.clip(depth_norm, 0, 1)
    if render_tab_state.inverse:
        depth_norm = 1 - depth_norm
    renders = apply_float_colormap(depth_norm, render_tab_state.colormap).cpu().numpy()
elif render_tab_state.render_mode == "normal":
    render_normals = render_normals[0] * 0.5 + 0.5  # normalize to [0, 1]
    renders = render_normals.cpu().numpy()
elif render_tab_state.render_mode == "alpha":
    alpha = render_alphas[0, ..., 0:1]
    renders = apply_float_colormap(alpha, render_tab_state.colormap).cpu().numpy()
```

This small change fixes the depth and normal rendering bugs.

Thanks again for the excellent repo!

PR #724, opened 2025-06-10
Adaptive MAX_JOBS depending on cpu count

atonderski:
Adaptively grow MAX_JOBS beyond 10 for systems with a larger number of cores.
This is purely for convenience, since MAX_JOBS can be set in the environment, but
probably a better default than a single value across all systems. 

For example, with a 9950x3d this reduces compile time from 45s to 30s.


liruilong940607:
I'm deliberately not setting it to cpu_count because on curtain systems this will crash the build (maybe bc of small RAM?) so 10 is a conservative setting here to make sure it does not crash.

atonderski:
aaah got it!

PR #734, opened 2025-06-22
fix 3dgut sample viewer

inuex35:
Clicking the top-right UI button in examples/simple_viewer_3dgut.py caused the viewer to crash due to an AssertionError related to the shape of thin_prism_coeffs:

`AssertionError: torch.Size([1, 4])
`

image

The issue was that thin_prism_coeffs was not shaped as expected by the renderer ([1, C, 2]).
To fix this, the tensor was redefined as:

```
thin_prism_coeffs = torch.tensor(
    [
        [
            gui_slider_thin_prism_coeffs1.value,  # s1
            gui_slider_thin_prism_coeffs2.value,  # s2
        ]
    ],
    device=device,
)
```

This resolves the crash and the viewer now works as expected.
It’s unclear whether this bug affects all users or only specific environments.

### DIFF
--- a/examples/simple_viewer_3dgut.py
+++ b/examples/simple_viewer_3dgut.py
@@ -167,8 +167,6 @@ def main(local_rank: int, world_rank, world_size: int, args):
                     [
                         gui_slider_thin_prism_coeffs1.value,  # s1
                         gui_slider_thin_prism_coeffs2.value,  # s2
-                        0.0,  # s3
-                        0.0,  # s4
                     ]
                 ],
                 device=device,


liruilong940607:
Actually it's the assert that is wrong. The `thin_prism_coeffs` should with shape [..., 4]

I also noticed there are a couple places in the code (both cuda and python) has wrong comments, like this by searching  `thin_prism_coeffs` (cursor is not doing a great job lol):

https://github.com/nerfstudio-project/gsplat/blob/0b4dddf04cb687367602c01196913cde6a743d70/gsplat/cuda/csrc/Projection.h#L273

Also the code here is wrong as well:

https://github.com/nerfstudio-project/gsplat/blob/0b4dddf04cb687367602c01196913cde6a743d70/gsplat/cuda/csrc/ProjectionUT3DGSFused.cu#L113

should shift by 4 instead of 2.

If you could fix them in this PR that would be great. Otherwise I will find some time to fix.

inuex35:
I updated the branch, could you review it?

liruilong940607:
Yeah this looks good. Since you are fixing it, would you mind fixing all comments at once in this PR?

You can search for `thin_prism_coeffs` in the code base and see where it has  comments like`# [..., C, 2] optional`

https://github.com/search?q=repo%3Anerfstudio-project%2Fgsplat%20thin_prism_coeffs&type=code

(make sure the expand the search results with "show x more matches")

liruilong940607:
LGMT now! thanks for the fix!

inuex35:
Thank you!

PR #735, opened 2025-06-22
fix: solves issue, that opacity is reset at iteration 0

kleinicke:
When the iterations start at 0, the default strategy resets the opacity after iteration 0.
The reset opacity is 2*prune opacity and not the init opacity.
Therefore the init opacity was probably unwillingly ignored.

This fix could lead to unexpected changes for some people, when suddenly the init opacity value is used instead of 2 times the prune opacity.


liruilong940607:
this make sense! merging it!

PR #743, opened 2025-06-27
Remove unnecessary abs() calls in scale/opacity loss calculation

vincentwoo:
the outputs of torch.sigmoid and torch.exp are always positive, therefore we need not call torch.abs on their results.


liruilong940607 (approved):
LGTM

Issue #746, opened 2025-07-02
GLM error for fresh gsplat instalation


randomze:
Hey, thanks for the great package!

When doing a fresh install of gsplat from PyPI, through `pip install gsplat`, I'm getting nvcc errors when it attempts to JIT compile the kernels. The error log is:

```
/home/ze/semmap/env_test/lib/python3.13/site-packages/torch/utils/cpp_extension.py:2356: UserWarning: TORCH_CUDA_ARCH_LIST is not set, all archs for visible cards are included for compilation.
If this is not desired, please set os.environ['TORCH_CUDA_ARCH_LIST'].
  warnings.warn(
freezing-myth :   0%|                                                                                                                                                                                                                                    | 0/30000 [00:04<?, ?it/s]
Traceback (most recent call last):
  File "/home/ze/semmap/env_test/lib/python3.13/site-packages/gsplat/cuda/_backend.py", line 166, in <module>
    from gsplat import csrc as _C
ImportError: cannot import name 'csrc' from 'gsplat' (/home/ze/semmap/env_test/lib/python3.13/site-packages/gsplat/__init__.py)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ze/semmap/env_test/lib/python3.13/site-packages/gsplat/cuda/_backend.py", line 79, in load_extension
    compiled = _jit_compile(
        name,
    ...<10 lines>...
        keep_intermediates=True,
    )
TypeError: _jit_compile() missing 2 required positional arguments: 'verbose' and 'with_sycl'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ze/semmap/env_test/lib/python3.13/site-packages/torch/utils/cpp_extension.py", line 2506, in _run_ninja_build
    subprocess.run(
    ~~~~~~~~~~~~~~^
        command,
        ^^^^^^^^
    ...<3 lines>...
        check=True,
        ^^^^^^^^^^^
        env=env)
        ^^^^^^^^
  File "/home/ze/.local/share/pyenv/versions/3.13.2/lib/python3.13/subprocess.py", line 579, in run
    raise CalledProcessError(retcode, process.args,
                             output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['ninja', '-v', '-j', '10']' returned non-zero exit status 1.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
[...]

PR #751, opened 2025-07-04
Fix lack of submodules in CI/CD

randomze:
Currently, the CI/CD pipeline uses GitHub Actions to produce the distribution packages. Except for the wheels, the remaining deployment paths use the `checkout@v4` action without `submodules: recursive`, causing GLM to not be present in the generated packages. This causes issue #746.

This pull request fixes the CI/CD configuration. New packages need to be published, especially since the default target from PiPY (gsplat==1.5.2), that you get if you run `pip install gsplat`, is broken right now (#746).

Hope this is enough to fix the issue and let me know if I need to do something more!

### DIFF
--- a/.github/workflows/building.yml
+++ b/.github/workflows/building.yml
@@ -7,6 +7,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: recursive
       - name: Strip unsupported tags in README
         run: |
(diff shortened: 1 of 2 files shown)


liruilong940607 (approved):
LGTM!

liruilong940607:
Thanks for the fix! v3 used to have recursive as default but I didn't realize v4 changed this behavior...

randomze:
Sneaky changes like this are very annoying. Glad to be able to give back to such a great project, keep up the good work 😁!

PR #798, opened 2025-09-13
Add support for AMD GPU

### LINKED ISSUE #434: Looking for an alternative with AMD GPU
Hi, is it possible to run gsplat on AMD GPUs? Are there any plans to support other languages like OpenCL?

### LINKED ISSUE #771: Plans on making GSplat compatible with AMD GPU

pxl-th:
This PR introduces AMD GPU support for gsplat.

Tested on:
- CPU: AMD Ryzen 7 5800X 8-Core @ 16x 4.594GHz
- GPU: Radeon RX 7900 XTX
- OS: Ubuntu 24.04
- ROCm: 6.4 (from [official installation script](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/quick-start.html#rocm-installation)).

Tested and confirmed working:
- 3DGS (simple & mcmc; packed & un-packed)
- 2DGS (packed & un-packed)
- 3DGUT

Relies on:
- https://github.com/rahul-goel/fused-ssim/pull/24
- https://github.com/harry7557558/fused-bilagrid/pull/6

Closes #771.
Closes #434.

image

- Move glm library out of the `cuda/` directory to avoid hipifying it, which causes confusion during compilation.
- Since AMD GPU does not support `cg::labeled_partition`, simply avoid warp reductions on it and just do global memory writes directly. To reduce code duplication, introduce `FOR_HIP` variable that we use to determine whether to use labeled partition or just a placeholder to avoid compilation errors, e.g.:
```cuda
#if FOR_HIP
auto warp_group_g = warp; // Not used, just here to not error in the if-statements.
#else
auto warp_group_g = cg::labeled_partition(warp, gid);
#endif
```
and in places where we'd do only one global atomic add (which should eliminate branching altogether):
```cuda
if (FOR_HIP || warp_group_g.thread_rank() == 0) {
...
}
```

- Replace `std::array.at` with `[]` indexing to avoid bounds-checking which causes errors:
```cuda
/usr/lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/array:220:9: error: reference to __host__ function '__throw_out_of_range_fmt' in __host__ __device__ function
  220 |           std::__throw_out_of_range_fmt(__N("array::at: __n (which is %zu) "
      |                ^
```

- Since ROCm does not support `cg::reduce`, create equivalent `warpSum` reduction methods using `shfl_down` intrinsic & use it where we do reduction on the whole warp (i.e. `tiled_partitiion` in our case).
- Use respective NVCC flags depending on the GPU backend. E.g. `-munsafe-fp-atomics` is required to replace CAS-loop with fast hardware floating-point atomics that significantly improves the performance on AMD GPU.


charyang-ai:
how is the performance comparied to Nvidia RTX card, like 4090?

Spacefish:
https://github.com/Spacefish/gsplat-rdna/tree/rocm_7_2 haha did the same thing but on AMDs fork of gsplat.

Performance is ok, i guess 8x8 is impacted a littlebit as we have to use shared memory to accumulate 2 32 wide warps but other sizes are fine / work on 32 wide waves like a charm.

charyang-ai:
@pxl-th  this repo works fine on RDNA4:  https://github.com/charyang-ai/gsplat-rocm-rdna4/tree/main

PR #831, opened 2025-11-13
Packed backward pass speedup via unrolled camera position indexing

gurki-bajwa-ai:
## Description
When running 3DGS with `packed=True, pose_opt=True` and using a CUDA device, the computation time for [this](https://github.com/nerfstudio-project/gsplat/blob/65042cc501d1cdbefaf1d6f61a9a47575eec8c71/gsplat/rendering.py#L498) campos indexing operation in the backwards pass is huge. This is because the gradients of ALL dirs would need to be accumulated into the gradient of the camera positions (which are much smaller in number), in the backward pass. The pytorch backward cuda kernel for this operation is very expensive due to numerous GPU atomic operations.

This PR unrolls the indexing and relies on pytorch's broadcasting for faster backwards pass, while keeping the overall numerical calculation exactly the same.

## Training times
I ran the `examples/benchmarks/basic.sh` on RTX 3090. Major improvement in the training time when pose_opt and packed are true. No major performance effect in the other case.

### batch_size=1 max_steps=5000 packed=true pose_opt=true
Before this change
| Scene   | Train Time | Number of splats | PSNR  | SSIM |
| ------- | ---------- | ---------------- | ----- | ---- |
| garden  | 496.45s    | 2084780          | 25.25 | 0.77 |
| bicycle | 283.62s    | 2048744          | 23.27 | 0.60 |
| stump   | 212.59s    | 2540310          | 24.32 | 0.64 |
| bonsai  | 345.74s    | 927848           | 29.14 | 0.92 |
| counter | 370.20s    | 622163           | 26.96 | 0.88 |

After this change
| Scene   | Train Time | Number of splats | PSNR  | SSIM |
| ------- | ---------- | ---------------- | ----- | ---- |
| garden  | 114.48s    | 2088360          | 25.30 | 0.76 |
| bicycle | 95.26s     | 2026429          | 23.30 | 0.60 |
| stump   | 92.29s     | 2522318          | 24.32 | 0.64 |
| bonsai  | 99.4s      | 925365           | 29.22 | 0.92 |
| counter | 97.83s     | 622668           | 26.86 | 0.88 |

### batch_size=1 max_steps=5000 packed=false pose_opt=false
Before this change
| Scene   | Train Time | Number of splats | PSNR  | SSIM |
| ------- | ---------- | ---------------- | ----- | ---- |
| garden  | 119.15s    | 2076023          | 25.63 | 0.78 |
| bicycle | 96.80s     | 2043134          | 23.40 | 0.61 |
| stump   | 100.20s    | 2592472          | 24.51 | 0.65 |
| bonsai  | 99.70s     | 911051           | 29.48 | 0.93 |
| counter | 97.50s     | 631954           | 26.95 | 0.88 |

After this change
| Scene   | Train Time | Number of splats | PSNR  | SSIM |
| ------- | ---------- | ---------------- | ----- | ---- |
| garden  | 117.12s    | 2069605          | 25.64 | 0.78 |
| bicycle | 96.47s     | 2047385          | 23.40 | 0.61 |
| stump   | 98.24s     | 2522734          | 24.53 | 0.65 |
| bonsai  | 98.94s     | 917302           | 29.42 | 0.93 |
| counter | 96.19s     | 631444           | 26.94 | 0.88 |

### DIFF
--- a/gsplat/rendering.py
+++ b/gsplat/rendering.py
@@ -67,7 +67,7 @@ def _compute_view_dirs_packed(
     campos_flat = campos.view(B, C, 3)
 
     if B * C == 1:
-        # Single batch-camera pair, no indexed lookup for camposneeded
+        # Single batch-camera pair. No indexed lookup for campos is needed.
         dirs = means_flat[0, gaussian_ids] - campos_flat[0, 0]  # [nnz, 3]
     else:
         avg_means_per_camera = nnz / (B * C)


gurki-bajwa-ai:
@liruilong940607 Please take a look and let me know if the PR text needs some theoretical justification too.

liruilong940607:
This is a great finding! But I'm concerned that the current way of writing it (loop over B and C) might leads to very slow speed when B and C is large. Ideally we should vectorize the compute there.

gsplat/rendering.py
```diff
@@ -493,10 +494,27 @@ def reshape_view(C: int, world_view: torch.Tensor, N_world: list) -> torch.Tenso
...
-            )  # [nnz, 3]
+            nnz = batch_ids.shape[0]
+            means_flat = means.view(B, N, 3)
+            campos_flat = campos.view(B, C, 3)
+
+            # Compute dirs in B*C steps to avoid many-to-one indexing of cameras in backward pass
+            dirs = torch.empty((nnz, 3), dtype=means.dtype, device=device)
+            indptr_cpu = indptr.cpu()
+            for b_idx in range(B):
+                for c_idx in range(C):
```
gurki-bajwa-ai: @liruilong940607 Good point about `B*C`. In most cases `B*C` will be much smaller than N, so the overhead should be low. Additionally each chunk of calculation within the loop is still vectorized so we should be fine on cpu. For cuda device, I understand your concern since there can be low GPU utilization if the numbers being processed in the loop are small.

I can address this by using this unrolling only if `nnz/(B*C)>10k` so that there is enough GPU occupancy. 10k is a reasonable number since modern desktop GPUs have 10k-20k cuda cores. Are you okay with that?

liruilong940607: gsplat is also used in certain ML cases (such as simulation) which have a large batch size / number of views with a small number of GS, so I need to take those cases into account.

But I think your suggestion makes sense. For cases B*C is small I think we could dispatch the impl to yours. How about we wrap this part into a standalone function of creating dirs, if which we could do an if-else? Just that I dont want this rasterization function to grow in line of code anymore it's already quite huge.

BTW `indptr_cpu = indptr.cpu()` this will incurr a GPU/CPU sync which is avoidable in the case if B=1 and C=1, in which `gids==gaussian_ids` and `b_idx==0`. Let's try to make this optimal :)

liruilong940607:
btw plz run the formatter to pass the test `black . gsplat/ tests/ examples/ profiling`

gurki-bajwa-ai:
I was using python3.10's formatter and hence failed the github tests. Used python3.8's formatter now. It should pass now.

liruilong940607:
Merging this now! Thank you for looking into this.

abhitoronto:
@gurki-bajwa-ai 
Could we possibly have optimized by flattening the matrices and then using highly optimized indexing operation?

```python
# 1. Flatten the source tensors
means_flat = means.view(-1, 3)   # Shape: (B*N, 3)
campos_flat = campos.view(-1, 3) # Shape: (B*C, 3)

# 2. Compute flat indices 
mean_indices = batch_ids * N + gaussian_ids
cam_indices = batch_ids * C + camera_ids

# 3. Fast indexing on flattened matrices
means_selected = torch.index_select(means.view(-1, 3), 0, mean_indices)
campos_selected = torch.index_select(campos.view(-1, 3), 0, cam_indices)

# 4. Difference
dirs = means_selected - campos_selected
```

PR #880, opened 2026-03-11
[NV] Fix wheel build

rodlima78:


.github/workflows/building.yml
```diff
@@ -135,7 +88,8 @@ jobs:
 
       - name: Upgrade pip
         run: |
-          pip install --upgrade setuptools
+          # https://github.com/pypa/setuptools/issues/5174, pkg_resources is removed in setuptools 82 but we need it for older torch versions
+          pip install --upgrade "setuptools<82"
```
liruilong940607: I think we probably want to make the build works for `setuptools>=82` as well as the user end might have `setuptools>=82` and it might just triggers fail building. Adding this restriction in the CI workflow feels escaping from the problem instead of solving it.

gsplat/cuda/include/ExternalDistortion.cuh
```diff
@@ -28,7 +28,7 @@
 namespace gsplat::extdist
 {
 
-class BivariateWindshieldModelDeviceParams;
+struct BivariateWindshieldModelDeviceParams;
```
vince-brisebois: Is this necessary, same with Lidars.cuh below. Revert?

PR #887, opened 2026-03-13
Integrating rmbrualla's pycolmap into gsplat

ichsan2895:
Compatibilty with pycolmap >=3.13

As `requirement.txt`, the pycolmap version that compatible with `simple_trainer.py` is `git+https://github.com/rmbrualla/pycolmap@cc7ea4b7301720ac29287dbe450952511b32125e`

When I use latest stable pycolmap (3.13.0), it gives error:
```
$ CUDA_VISIBLE_DEVICES=0 python3 gsplat-dev/examples/simple_trainer.py mcmc --help
  File "/GSPLAT_152n3/gsplat-dev/examples/simple_trainer.py", line 18, in <module>
    from datasets.colmap import Dataset, Parser
  File "/GSPLAT_152n3/gsplat-dev/examples/datasets/colmap.py", line 10, in <module>
    from pycolmap import SceneManager
ImportError: cannot import name 'SceneManager' from 'pycolmap' (/usr/local/lib/python3.10/dist-packages/pycolmap/__init__.py)
```

Now this PR makes the bug fixes.


ichsan2895:
Strange, I have run black formatting, but it still fails.
image

######
Hmm the sentence, "adding pycolmap >=3.13" might be confusing. How about changing the title "Integrating rmbrualla's pycolmap into gsplat"

`gsplat simple_traineer.py` needs special `pycolmap` called `rmbrualla pycolmap`. If I install `rmbrualla's pycolmap`, it uninstall the existing pycolmap (if user has it).

So, the PR makes sure it wont break each others (rmbrualla's pycolmap does not need to be installed) since it integrated with gsplat

PR #891, opened 2026-03-14
Fix #444: Add pre-allocation option for MCMC strategy to prevent OOM

### LINKED ISSUE #444: MCMC memory consumption
I've been using gsplat’s MCMC strategy for high-quality reconstructions but have encountered some scalability limitations. Specifically, on my 48GB GPU, I’m having trouble refining the model beyond approximately 18 million points due to VRAM exhaustion. Interestingly, if I start with a random initialization of 30 million points, I can achieve stable training over many iterations, but the reconstruction quality is noticeably degraded, as expected. 

From my observations, I suspect that the root of this issue may be related to the use of `torch.cat` in functions like [sample_add](https://github.com/nerfstudio-project/gsplat/blob/d4020bcc869b9390ccb3f35906056640e7cf208a/gsplat/strategy/ops.py#L295). For instance, to extend the solution to handle N+M points, it seems the current approach requires not only the N original points and the M new points but also a contiguous buffer of size N+M to perform the concatenation. For reference, at 18 million points, this already demands about 4.3GB of memory. Beyond that, additional memory is required for gradients and Adam optimizer’s moving averages, and the capacity of these buffers also needs to be increased.

I’m wondering if there are potential optimizations to reduce this memory overhead. Would it be feasible to allocate a fixed buffer of size `cap-max`, tracking which of the first N elements in the tensor are active? This might reduce temporary storage for concatenations and help stabilize VRAM usage throughout training, and might be easier to accomplish due to the recent addition of SelectiveAdam.

Perhaps there any related features or configuration settings I have missed? I’m already using `--sparse-grad`, which has been very helpful for large scenes. I’ve also tried tweaking the `PYTORCH_CUDA_ALLOC_CONF` environment variable, but without much success.

eladerez:
The problem
MCMC training crashes with OOM errors at ~18M Gaussians due to memory fragmentation from repeated torch.cat operations during the densification stage.

Solution:
Added an optional "preallocate" flag to "MCMCStrategy" that:
Pre-allocates buffers to "cap_max" size at initialization
Writing new Gaussians into pre-allocated slots instead of concatenating

Changes:
gsplat/strategy/mcmc.py: Added "preallocate" field and "n_active" state tracking
gsplat/strategy/ops.py: Modified sample_add() for support to buffer writes
examples/simple_trainer.py: Managing buffer and checkpoint

Performance:
Memory: 6.4% reduction at 1M Gaussians (1.216GB vs 1.299GB baseline)
Trade-off: ~24% slower rendering (opt-in feature, disabled by default)
Benefit: Prevents OOM at 18M+ Gaussians scale

Testing:
All strategy tests passing (2/2)

How to run:
python examples/simple_trainer.py mcmc --strategy.preallocate

### DIFF
--- a/examples/simple_trainer.py
+++ b/examples/simple_trainer.py
@@ -484,12 +484,31 @@ class Runner:
                 scene_scale=self.scene_scale
             )
         elif isinstance(self.cfg.strategy, MCMCStrategy):
-            # We pass the initial active count when preallocating
             if cfg.strategy.preallocate:
                 n_initial = len(self.parser.points) // world_size
                 self.strategy_state = self.cfg.strategy.initialize_state(
                     n_initial=n_initial
                 )
+                # Build active-slice params (views into the pre-allocated buffers)
+                # and rewire each optimizer to track the active slice instead of the
+                # full cap_max-sized tensor so that optimizer.step() only touches
+                # the n_initial live Gaussians at startup.
+                active_params = {
+                    name: torch.nn.Parameter(
+                        self.splats[name].data[:n_initial],
+                        requires_grad=self.splats[name].requires_grad,
+                    )
+                    for name in self.splats.keys()
+                }
+                for name, opt in self.optimizers.items():
+                    if name not in active_params:
+                        continue
+                    full_param = self.splats[name]
+                    for group in opt.param_groups:
+                        for i, p in enumerate(group["params"]):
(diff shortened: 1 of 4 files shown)


eladerez:
I've submitted PR #891 that addresses this with pre-allocated buffers. Would appreciate feedback.

PR #893, opened 2026-03-16
Add scipy as a required dependency

andmax:
Python 3.8 is EOL (Oct 2024) and incompatible with the codebase: dataclass(kw_only=True) requires Python 3.10+.  Also bump actions/checkout to v4 to fix Node.js deprecation warning.

The lidar module (gsplat/cuda/_lidar.py) imports scipy at module level for cKDTree.  Since _lidar.py is imported transitively on every 'import gsplat', scipy must be installed.

### DIFF
--- a/.github/workflows/core_tests.yml
+++ b/.github/workflows/core_tests.yml
@@ -14,14 +14,14 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           submodules: 'recursive'
-          
-      - name: Set up Python 3.8.12
+
+      - name: Set up Python 3.10
         uses: actions/setup-python@v5
         with:
-          python-version: "3.8.12"
+          python-version: "3.10"
       - name: Install dependencies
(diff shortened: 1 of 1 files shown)


rodlima78:
Ok, changes approved, thanks.

ndeybach:
Hi, thanks for the renewed work in gsplat these past few weeks and merged PRs again !

Regarding this PR, would it be possible to rather do a weak import in the lidar module (like in the function where it is used, so that it loads scipy module only the first time `compute_angles_to_columns_map` is called ?).

Since if done at the top of file/module and loaded at any `import gsplat` it is pretty much dead load time for anyone (scipy is pretty heavy library) even though not everyone uses the lidar functionality ? 

Scipy could also be an optional dependency since forcing an installation only for a single, non essential function, seems a bit of a heavy approach ?

rodlima78:
Thanks for you message, @ndeybach . Indeed it's only needed for lidar use cases. But I feel that scipy should be specified at least as an optional dependency at build/install time, and go to the weak import path. We'll come up with a PR to address this issue.

PR #930, opened 2026-04-22
Added median depth rasterization for 3dgs [fwd only]

sangeethagramas:
Change : Add median depth rasterization for 3dgs  for the forward pass only

Description : This commit currently supports median depth calculation for 3D gaussians only on the forward pass. We currently do not anticipate the need for median depth support in the backward pass. We can add it in the future if the need arises.
The changes in this PR follows the style similar to 2D gaussian median depth rasterization.

Test cases :
1. Group 1: sanity check tests for output shape checks on sample scene
2. Group 2: check workflow for depth_mode=="median"
3. Group 3: sanity check for API validation (rejects undefined/ill-specified inputs)
4. Group 4: median depth kernel tests for dummy scenes


gsplat/cuda/csrc/Rasterization.cpp
```diff
@@ -101,7 +110,9 @@ std::tuple<at::Tensor, at::Tensor, at::Tensor> rasterize_to_pixels_3dgs_fwd(
             flatten_ids,                                                       \
             renders,                                                           \
             alphas,                                                            \
-            last_ids                                                           \
+            last_ids,                                                          \
+            render_median,                                                     \
+            median_ids                                                         \
```
liruilong940607, line 115: I think we should make them optional tensors -- to make sure we dont introduce any perf regression when median depth is not needed.

gsplat/cuda/csrc/RasterizeToPixels3DGSFwd.cu
```diff
@@ -59,7 +59,13 @@ __global__ void rasterize_to_pixels_3dgs_fwd_kernel(
...
         *__restrict__ render_colors, // [I, image_height, image_width, CDIM]
     scalar_t *__restrict__ render_alphas, // [I, image_height, image_width, 1]
-    int32_t *__restrict__ last_ids        // [I, image_height, image_width]
+    int32_t *__restrict__ last_ids,       // [I, image_height, image_width]
+    scalar_t
+        *__restrict__ render_median, // [I, image_height, image_width, 1]
+                                     // depth of the Gaussian that causes
+                                     // transmittance to cross 0.5, or the
+                                     // last-hit depth as a fallback.
+    int32_t *__restrict__ median_ids // [I, image_height, image_width]
```
liruilong940607, line 68: related to optional: we want to allow them to be null ptr when not needed. and in the function we skip compute median depth and write out so we dont waste compute if they are not needed

gsplat/cuda/csrc/RasterizeToPixels3DGSFwd.cu
```diff
@@ -190,6 +206,17 @@ __global__ void rasterize_to_pixels_3dgs_fwd_kernel(
             }
             cur_idx = batch_start + t;
 
+            // Track last hit depth (depth is stored as the last channel of
+            // colors, consistent with the existing D/ED plumbing and 2DGS).
+            const float dep = c_ptr[CDIM - 1];
```
liruilong940607, line 211: This silently assume render mode is in ED or D. Should assert this to protect in the rendering.py

liruilong940607: Also for the best performance, I think you can push this link and next into the `if (!median_found && T > 0.5f && next_T <= 0.5f) {` condition?

gsplat/cuda/csrc/RasterizeToPixels3DGSFwd.cu
```diff
@@ -209,6 +236,14 @@ __global__ void rasterize_to_pixels_3dgs_fwd_kernel(
...
         // index in bin of last gaussian in this pixel
         last_ids[pix_id] = static_cast<int32_t>(cur_idx);
+
+        // Median depth: if transmittance never crossed 0.5 fall back to the
+        // last contributing Gaussian's depth (0.0 if there were no hits).
+        if (!median_found) {
+            median_depth = last_hit_depth;
+        }
+        render_median[pix_id] = median_depth;
+        median_ids[pix_id] = static_cast<int32_t>(median_idx);
```
liruilong940607, line 246: Skip write if median is not required (`render_median` is a null ptr)

gsplat/cuda/csrc/RasterizeToPixels3DGSFwd.cu
```diff
@@ -229,9 +264,11 @@ void launch_rasterize_to_pixels_3dgs_fwd_kernel(
     const at::Tensor tile_offsets, // [..., tile_height, tile_width]
     const at::Tensor flatten_ids,  // [n_isects]
     // outputs
-    at::Tensor renders, // [..., image_height, image_width, channels]
-    at::Tensor alphas,  // [..., image_height, image_width]
-    at::Tensor last_ids // [..., image_height, image_width]
+    at::Tensor renders,       // [..., image_height, image_width, channels]
+    at::Tensor alphas,        // [..., image_height, image_width]
+    at::Tensor last_ids,      // [..., image_height, image_width]
+    at::Tensor render_median, // [..., image_height, image_width, 1]
```
liruilong940607, line 270: to be consistent on the naming convension in gsplat, suggested naming is `median_depths`

gsplat/cuda/csrc/RasterizeToPixels3DGSFwd.cu
```diff
@@ -286,7 +323,9 @@ void launch_rasterize_to_pixels_3dgs_fwd_kernel(
             flatten_ids.data_ptr<int32_t>(),
             renders.data_ptr<float>(),
             alphas.data_ptr<float>(),
-            last_ids.data_ptr<int32_t>()
+            last_ids.data_ptr<int32_t>(),
+            render_median.data_ptr<float>(),
+            median_ids.data_ptr<int32_t>()
```
liruilong940607, line 328: update for optional

gsplat/cuda/_wrapper.py
```diff
@@ -857,12 +858,20 @@ def rasterize_to_pixels(
         masks: Optional tile mask to skip rendering GS to masked tiles. [..., tile_height, tile_width]. Default: None.
         packed: If True, the input tensors are expected to be packed with shape [nnz, ...]. Default: False.
         absgrad: If True, the backward pass will compute a `.absgrad` attribute for `means2d`. Default: False.
+        return_median: If True, additionally returns per-pixel median depth, defined
+            as the value of the last channel of ``colors`` for the Gaussian whose
```
liruilong940607, line 862: "the value of the last channel of ``colors`` " is weird to read. Maybe just say something like "expect the last channel to be depth"

gsplat/cuda/_wrapper.py
```diff
@@ -941,7 +950,7 @@ def rasterize_to_pixels(
         tile_width * tile_size >= image_width
     ), f"Assert Failed: {tile_width} * {tile_size} >= {image_width}"
 
-    render_colors, render_alphas = _RasterizeToPixels.apply(
+    render_colors, render_alphas, render_median, _median_ids = _RasterizeToPixels.apply(
```
liruilong940607, line 953: if we are not returning `median ids` we might just remove it entirely?

gsplat/rendering.py
```diff
@@ -308,6 +308,7 @@ def rasterization(
     extra_signals_sh_degree: Optional[
         int
     ] = None,  # Currently only None or 3 is accepted.
+    depth_mode: Literal["expected", "median"] = "expected",
```
liruilong940607, line 311: "depth_mode" is a bit weird here because we already say render mode has ED or D. this "depth mode" between expected and median could be very confusing on the relationship between the ED and D in render mode.

My suggested interface change:

use argument `return_median_depth: bool = False` here. Then additionally return median depth (and id if you need it) if this is True. If not only return the original returns, and do not waste any compute in the kernel for median depth. This way we are backward compatible and cause minimal confusion.

PR #937, opened 2026-04-30
fix(docs): bump Python to 3.10 in docs CI to fix empty API pages

vince-brisebois:
Root cause: The docs CI uses Python 3.9, which caps torch at ~2.6. These older torch versions are incompatible with modern setuptools>=82 (which removed pkg_resources), causing import torch.utils.cpp_extension to fail. Since   _backend.py is imported at module level via _make_lazy_cuda_cls() in _wrapper.py, this crashes the entire import gsplat chain. Sphinx catches the error silently and deploys empty pages.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -21,7 +21,7 @@ jobs:
           
       - uses: actions/setup-python@v3
         with:
-          python-version: '3.9'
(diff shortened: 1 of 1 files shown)


liruilong940607 (approved):
LGTM!

PR #938, opened 2026-05-04
fix: relax backgrounds shape assertion when packed=True

### LINKED ISSUE #826: assert on  backgrounds.shape seems wrong in function "rasterize_to_pixels" when packed is True
Assert on backgrounds.shape seems wrong when packed is True. When viewmats shape is C x 4 x 4, backgrounds should always be C x D since backgrounds for each image could be different. The corresponding code is here: https://github.com/nerfstudio-project/gsplat/blob/65042cc501d1cdbefaf1d6f61a9a47575eec8c71/gsplat/cuda/_wrapper.py#L598

jashshah999:
## What

When `packed=True`, `means2d.shape` is `(nnz, 2)` so `image_dims = means2d.shape[:-2]` is `()`. The assertion `backgrounds.shape == () + (channels,)` then requires backgrounds to be `(channels,)`, but it should be `(C, channels)` where `C` is the number of cameras (each image can have a different background).

## Fix

When packed, only check that the last dimension of backgrounds matches `channels`. The batch dimensions vary depending on the number of cameras.

Fixes #826.

### DIFF
--- a/gsplat/cuda/_wrapper.py
+++ b/gsplat/cuda/_wrapper.py
@@ -881,7 +881,14 @@ def rasterize_to_pixels(
         assert colors.shape == image_dims + (N, channels), colors.shape
         assert opacities.shape == image_dims + (N,), opacities.shape
     if backgrounds is not None:
-        assert backgrounds.shape == image_dims + (channels,), backgrounds.shape
+        if packed:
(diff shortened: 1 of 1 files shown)


vince-brisebois (commented):
(removed)

PR #941, opened 2026-05-04
feat: add AdaptiveStrategy for auto-tuned densification

jashshah999:
## What

New densification strategy that eliminates manual threshold tuning. Instead of requiring users to set `grow_grad2d`, `prune_opa`, `grow_scale3d`, etc., AdaptiveStrategy uses a target growth rate and lets the gradient distribution determine what gets densified.

## Key Idea

Instead of: "densify anything with gradient > 0.0002" (needs per-scene tuning)

Use: "densify enough Gaussians to grow by 50% each refine step" (scene-agnostic)

The gradient threshold is computed dynamically to select the right number of Gaussians. Combined with `cap_max`, this self-regulates: grows aggressively early when the scene is under-represented, then slows naturally as it converges.

## Benchmark

Synthetic 8-view scene, 3000 steps, same initial point cloud:

| Strategy | PSNR | #Gaussians | Tuning Required |
|----------|------|-----------|----------------|
| DefaultStrategy (hand-tuned) | 34.11 dB | 125k | grow_grad2d, prune_opa, grow_scale3d, prune_scale3d, ... |
| **AdaptiveStrategy** | **34.01 dB** | 190k | grow_rate, cap_max |

Same quality, zero threshold tuning.

## Usage

```python
from gsplat import AdaptiveStrategy

strategy = AdaptiveStrategy(
    grow_rate=0.5,       # grow by 50% each refine step
    cap_max=500_000,     # stop growing at 500k Gaussians
    prune_percentile=1.0,  # prune bottom 1% by opacity
)
# Same API as DefaultStrategy otherwise
```

## How it self-regulates

1. Early training: high gradients everywhere, growth rate is achieved easily
2. Mid training: gradients decrease, threshold drops to maintain growth rate
3. At cap_max: growth equals pruning, system reaches equilibrium
4. Late training (after refine_stop_iter): no more densification, just optimization

## API

Same interface as DefaultStrategy. Only addition: pass `loss=loss.item()` to `step_post_backward` for optional plateau detection.

Includes benchmark script (`benchmark_adaptive.py`).

### DIFF
--- a/benchmark_adaptive.py
+++ b/benchmark_adaptive.py
@@ -1,0 +1,225 @@
+"""Benchmark AdaptiveStrategy vs DefaultStrategy on a synthetic scene.
+
+Generates a test scene, trains with both strategies, compares PSNR.
+"""
+
+import time
+import torch
+import numpy as np
+from torch import Tensor
+from gsplat import rasterization
+from gsplat.strategy import DefaultStrategy
+
+# Import our adaptive strategy from local
+import sys
+sys.path.insert(0, ".")
+from gsplat.strategy.adaptive import AdaptiveStrategy
+
+
+def generate_test_scene(n_points=5000, device="cuda"):
+    """Generate a random colored point cloud as initial Gaussians."""
+    torch.manual_seed(42)
+    means = torch.randn(n_points, 3, device=device) * 2.0
+    colors = torch.rand(n_points, 3, device=device)
+    scales = torch.full((n_points, 3), -3.0, device=device)  # log scale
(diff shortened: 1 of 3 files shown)


jashshah999:
Fixed. The issue was using a static percentile (which gives too few Gaussians early in training). Switched to a target-growth-rate approach: the strategy computes the gradient threshold dynamically to achieve grow_rate fraction of growth per refine step. cap_max prevents runaway.

Results on synthetic scene: 34.01 dB vs DefaultStrategy's 34.11 dB, with zero threshold tuning. Updated the PR with the fix and benchmarks.

PR #953, opened 2026-05-20
Update RelocationCUDA.cu: fix mcmc relocation_kernel: avoid pow(-1.0f, k) return "Nan"

DrShawnDai:
When I was using gsplat's MCMC strategy, I found that around line 58 in RelocationCUDA.cu calculates the new scale using `pow(-1.0f, k)`. In my case, if `-use_fast_math` is used when compiling gsplat, it may cause `pow(-1.0f, k)` to return `NaN`. So I think replacing the `pow(-1.0f, k)` calculation with `float sign = (k % 2 == 0) ? 1.0f : -1.0f` be safer and faster.

### DIFF
--- a/gsplat/cuda/csrc/RelocationCUDA.cu
+++ b/gsplat/cuda/csrc/RelocationCUDA.cu
@@ -55,7 +55,8 @@ __global__ void relocation_kernel(
     for (int i = 1; i <= n_idx; ++i) {
         for (int k = 0; k <= (i - 1); ++k) {
             float bin_coeff = binoms[(i - 1) * n_max + k];
-            float term = (pow(-1.0f, k) / sqrt(static_cast<float>(k + 1))) *
+            float sign = (k % 2 == 0) ? 1.0f : -1.0f;
+            float term = (sign / sqrt(static_cast<float>(k + 1))) *
                          pow(new_opacities[idx], k + 1);
             denom_sum += (bin_coeff * term);
         }


vince-brisebois:
Hi,
We're trying to increase our test coverage and have proposed tests to cover this feature. Please consider adding them, or otherwise create your own code to demonstrate the issue reported is actually resolved. Cheers!

gsplat/cuda/csrc/RelocationCUDA.cu
```diff
@@ -55,7 +55,8 @@ __global__ void relocation_kernel(
     for (int i = 1; i <= n_idx; ++i) {
         for (int k = 0; k <= (i - 1); ++k) {
             float bin_coeff = binoms[(i - 1) * n_max + k];
-            float term = (pow(-1.0f, k) / sqrt(static_cast<float>(k + 1))) *
+            float sign = (k % 2 == 0) ? 1.0f : -1.0f;
+            float term = (sign / sqrt(static_cast<float>(k + 1))) *
```
vince-brisebois, line 59: The relocation kernel has no test coverage anywhere in `tests/` (grep for `compute_relocation`, `relocation`, or `binoms` finds zero hits in the test suite). The bug being fixed here — `pow(-1.0f, k)` silently returning NaN, which then propagates into `denom_sum` and produces NaN `new_scales` — is exactly the kind of regression a single forward-pass test would catch. Without a test, a future refactor could reintroduce the NaN (or break sign correctness, since the new code path is also untested).

The sign-by-parity computation is also subtly different from `pow(-1.0f, k)`: it should produce identical results for non-negative integer `k`, but a test pinning the expected output against a PyTorch reference is the only way to verify behavioral equivalence.

**Suggested test** (in `tests/test_relocation.py`):

```python
# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
# SPDX-License-Identifier: Apache-2.0

"""Tests for the CUDA relocation kernel.

Verifies that gsplat.compute_relocation produces finite outputs (regression
test for the pow(-1.0f, k) -> NaN bug fixed in RelocationCUDA.cu) and that
the output matches a pure-PyTorch reference implementation of Equation (9)
from "3D Gaussian Splatting as Markov Chain Monte Carlo".

Usage:
    pytest tests/test_relocation.py -s
"""

import math

import pytest
import torch

import gsplat
from gsplat.relocation import compute_relocation

device = torch.device("cuda:0")


def _make_binoms(n_max: int) -> torch.Tensor:
    binoms = torch.zeros((n_max, n_max), device=device, dtype=torch.float32)
    for n in range(n_max):
        for k in range(n + 1):
            binoms[n, k] = math.comb(n, k)
    return binoms


def _pytorch_reference(opacities, scales, ratios, binoms, n_max):
    """Pure-PyTorch reference for Eq. (9) matching the CUDA kernel."""
    N = opacities.shape[0]
    new_opacities = 1.0 - (1.0 - opacities) ** (1.0 / ratios.float())
    new_scales = torch.empty_like(scales)
    for idx in range(N):
        n_idx = int(ratios[idx].item())
        denom_sum = 0.0
        for i in range(1, n_idx + 1):
            for k in range(i):
                sign = 1.0 if (k % 2 == 0) else -1.0
                bin_coeff = binoms[i - 1, k].item()
                term = (sign / math.sqrt(k + 1)) * (
                    new_opacities[idx].item() ** (k + 1)
                )
                denom_sum += bin_coeff * term
        coeff = opacities[idx].item() / denom_sum
        new_scales[idx] = coeff * scales[idx]
    return new_opacities, new_scales


@pytest.mark.skipif(not torch.cuda.is_available(), reason="No CUDA device")
@pytest.mark.skipif(
    not hasattr(torch.ops.gsplat, "relocation"),
    reason="relocation op not built",
)
class TestRelocation:
    def test_outputs_are_finite(self):
        """Regression test for the pow(-1.0f, k) -> NaN bug.

        Before the fix, denom_sum became NaN on affected platforms,
        producing NaN new_scales. Exercise a wide range of ratios so
        we hit large k values where the bug manifested.
        """
        torch.manual_seed(0)
        n_max = 51
        binoms = _make_binoms(n_max)
        N = 256
        opacities = torch.rand(N, device=device, dtype=torch.float32) * 0.9 + 0.05
        scales = torch.rand(N, 3, device=device, dtype=torch.float32) + 0.1
        # Cover small and large ratios (the bug was worse for large k)
        ratios = torch.randint(1, n_max + 1, (N,), device=device, dtype=torch.int32)

        new_opacities, new_scales = compute_relocation(
            opacities, scales, ratios, binoms
        )

        assert torch.isfinite(new_opacities).all(), "new_opacities contains NaN/Inf"
        assert torch.isfinite(new_scales).all(), "new_scales contains NaN/Inf"
        assert (new_opacities > 0).all() and (new_opacities <= 1).all()
        assert (new_scales > 0).all()

    def test_matches_pytorch_reference(self):
        """CUDA output should numerically match a pure-PyTorch reference."""
        torch.manual_seed(1)
        n_max = 51
        binoms = _make_binoms(n_max)
        N = 32
        opacities = torch.rand(N, device=device, dtype=torch.float32) * 0.8 + 0.1
        scales = torch.rand(N, 3, device=device, dtype=torch.float32) + 0.1
        ratios = torch.randint(1, n_max + 1, (N,), device=device, dtype=torch.int32)

        actual_op, actual_sc = compute_relocation(
            opacities.clone(), scales.clone(), ratios.clone(), binoms
        )
        expected_op, expected_sc = _pytorch_reference(
            opacities, scales, ratios, binoms, n_max
        )

        torch.testing.assert_close(actual_op, expected_op, rtol=1e-4, atol=1e-5)
        torch.testing.assert_close(actual_sc, expected_sc, rtol=1e-3, atol=1e-4)

    def test_ratio_one_is_identity_like(self):
        """Edge case: ratio==1 means n_idx==1, so the inner loop runs once
        with k==0 (sign==+1). new_opacities should equal opacities.
        """
        binoms = _make_binoms(51)
        opacities = torch.tensor([0.1, 0.5, 0.9], device=device, dtype=torch.float32)
        scales = torch.ones(3, 3, device=device, dtype=torch.float32)
        ratios = torch.ones(3, device=device, dtype=torch.int32)

        new_op, new_sc = compute_relocation(opacities, scales, ratios, binoms)

        torch.testing.assert_close(new_op, opacities, rtol=1e-5, atol=1e-6)
        assert torch.isfinite(new_sc).all()

    def test_empty_input(self):
        """Edge case: N==0 should be a no-op (the launcher early-returns)."""
        binoms = _make_binoms(51)
        opacities = torch.empty(0, device=device, dtype=torch.float32)
        scales = torch.empty(0, 3, device=device, dtype=torch.float32)
        ratios = torch.empty(0, device=device, dtype=torch.int32)

        new_op, new_sc = compute_relocation(opacities, scales, ratios, binoms)

        assert new_op.shape == (0,)
        assert new_sc.shape == (0, 3)
```

---
*Generated by a review agent (test coverage)*

PR #954, opened 2026-05-22
NHT Support

Arcanous98:
## Summary

This PR adds support for Neural Harmonic Textures (https://research.nvidia.com/labs/sil/projects/neural-harmonic-textures/)

### 1. Neural Harmonic Textures (NHT) support

NHT is an **opt-in** extension to the rasterization pipeline that stores per-Gaussian features at four tetrahedron vertices and decodes them into RGB (or arbitrary AOVs) through a deferred shader. NHT only activates when `nht_params=NHTParams(...)` is passed to `rasterization`; default gsplat behavior is unchanged.

Highlights:

- Forward / backward CUDA kernels for the NHT 3DGS path (`RasterizeToPixelsFromWorldNHT3DGS{Fwd,Bwd}.cu`, `Interpolation.cuh`, `HalfVectorLoads.cuh`, `CommonNHT.h`, `OpsNHT.h`, `RasterizationNHT.{cpp,h}`).
- Public Python API: `gsplat.nht.{NHTParams, DeferredShaderModule, DeferredShaderModuleAOV, HarmonicFeatures, export_splats_nht}`.
- Routing in `rasterization(..., nht_params=...)` through the existing 3DGUT (`with_ut + with_eval3d`) projection in `gsplat/rendering.py`.
- `[nht]` extras in `setup.py` for the deferred-shader `tinycudann` dependency.
- New `examples/simple_trainer_nht.py`, `examples/simple_viewer_nht.py`, `examples/benchmarks/basic_nht.sh` and `examples/benchmarks/nht/`, plus NHT options in `profiling/{batch,main}.py`.
- `tests/test_nht.py` and full usage docs in `docs/nht.md`.
- [NEW] Extended fisheye / orthographic / lidar camera support in the NHT path.
- [NEW] Support for depth and normals rendering, and backward pass to obtain gradients to all primitive parameters to supervise on depth or normal signals.
- [NEW] Simple API to fit arbitrary signals through the deferred shading module (AOVs)

### 2. Viewer + trainer: orthographic camera and normals render modes

Independent of NHT, `examples/{simple_viewer,gsplat_viewer,simple_trainer}.py` gain:

- `apply_ortho_scale_to_K(K, w, h, scale)` helper to enable arbitrary scaling when using ortographic projections in the viewer.
- A new `"normal"` render-mode dropdown that visualizes depth-derived normals (`return_normals=True` with `with_eval3d=True`).

[BUG FIX] Lidar global_z_order fix

In `rendering.py`, lidar projection now enforces `global_z_order=False` to make sure culling is correct when using lidar cameras.

## Test plan

- [x] `BUILD_NHT=1 BUILD_3DGS=1 BUILD_3DGUT=1 pip install -e . --no-build-isolation` builds clean (Linux + Windows, CUDA 12.x).
- [x] `pytest tests/test_nht.py` passes.
- [x] Existing test suite still passes (no regressions in default code paths).
- [x] `python examples/simple_trainer_nht.py mcmc --help` runs.
- [x] `python examples/simple_viewer.py` renders RGB / depth / normal modes and the new `Ortho Scale` slider behaves correctly.
- [x] Test of the mipnerf360 benchmarking scripts under ./examples/benchmarks/nht


gsplat/cuda/include/Ops.h
```diff
@@ -686,3 +686,5 @@ void mcmc_perturb_positions(
 );
 
 } // namespace gsplat
+
+#include "OpsNHT.h"
```
wilsonCernWq: Do you think it would be better if we explicitly include NHT ops in only NHT related files?

gsplat/cuda/ext.cpp
```diff
@@ -767,6 +773,11 @@ TORCH_LIBRARY(gsplat, m) {
     m.def("mcmc_perturb_positions(Tensor(a!) positions, Tensor quats, Tensor scales, Tensor opacities, Tensor noise, float scaler) -> ()");
 #endif
 
+#if GSPLAT_BUILD_NHT && GSPLAT_BUILD_3DGS
+    m.def("rasterize_to_pixels_from_world_nht_3dgs_fwd(Tensor means, Tensor quats, Tensor scales, Tensor colors, Tensor opacities, Tensor? backgrounds, Tensor? masks, int image_width, int image_height, int tile_size, Tensor viewmats0, Tensor? viewmats1, Tensor Ks, int camera_model, __torch__.torch.classes.gsplat.UnscentedTransformParameters ut_params, int rs_type, Tensor? radial_coeffs, Tensor? tangential_coeffs, Tensor? thin_prism_coeffs, __torch__.torch.classes.gsplat.FThetaCameraDistortionParameters ftheta_coeffs, __torch__.torch.classes.gsplat.RowOffsetStructuredSpinningLidarModelParametersExt? lidar_coeffs, __torch__.torch.classes.gsplat.BivariateWindshieldModelParameters? external_distortion_params, Tensor tile_offsets, Tensor flatten_ids, bool center_ray_mode, float ray_dir_scale, Tensor? depths_per_gauss, bool use_hit_distance, bool with_normals) -> (Tensor, Tensor, Tensor, Tensor, Tensor)");
```
wilsonCernWq, line 1274: I am not an expert on this part, but why some has `Tensor?`? The question mark I meant.

setup.py
```diff
@@ -91,6 +91,12 @@ def get_extensions():
         "lidar": [
             "scipy",
         ],
+        # NHT (Neural Harmonic Textures) deferred-shader dependencies.
+        # Install them with `pip install "gsplat[nht]"`. tiny-cuda-nn requires
+        # a working CUDA toolchain at install time; see docs/nht.md.
+        "nht": [
+            "tinycudann @ git+https://github.com/NVlabs/tiny-cuda-nn/#subdirectory=bindings/torch",
```
wilsonCernWq: I think the default CI will not test NHT dependencies. Maybe let's add this test in our own landing repo's CI?

examples/gsplat_viewer.py
```diff
@@ -37,7 +37,7 @@ class GsplatRenderTabState(RenderTabState):
     eps2d: float = 0.3
     backgrounds: Tuple[float, float, float] = (0.0, 0.0, 0.0)
     render_mode: Literal[
-        "rgb", "depth(accumulated)", "depth(expected)", "alpha"
+        "rgb", "depth(accumulated)", "depth(expected)", "normal", "alpha"
```
wilsonCernWq, line 40: maybe we could add some extra docs here and in other similar places, reminding that this new mode is NHT related? I feel this would make future maintainer easier.

Arcanous98: Actually I added also depth derived normals for regular 3dgs gsplat. But I agree this is a separate issue, so we can potentially spin it out of this PR.

gsplat/cuda/ext.cpp
Arcanous98: It indicates that it is optional, it can be None depending on the rendering mode.

gsplat/cuda/include/Ops.h
Arcanous98: Maybe, I'll see what I can do to encapsulate this out of the global ops header 🙂

Arcanous98:
I cleaned the nht ops to be only loaded when instaling NHT, and added CI tests. Let me know if I can keep the normals visualization in the gsplat viewer or if I should postpone it, and if I can do something more to help make the MR ready

.github/workflows/nht_tests.yml
```diff
@@ -0,0 +1,109 @@
+name: NHT Tests
```
wilsonCernWq: let's move this test to https://github.com/nv-tlabs/neural-harmonic-textures/blob/main/.github/workflows/ci.yaml, to avoid adding workload to gsplat's general CI?

examples/gsplat_viewer.py
wilsonCernWq: I see, let's just document them well then.

Arcanous98:
I fixed some errors popping from the current gsplat CI workflows and moved the NHT ones to our old repository as discussed @wilsonCernWq :)

gsplat/cuda/include/ExternalDistortion.cuh
```diff
@@ -36,7 +36,10 @@ struct BivariateWindshieldModelDeviceParams;
 // - N CUDA architectures/compute capability targets (e.g. 80, 86, 89, 90, 100, 120)
 // - Inlining into all contexts creates O(n³) register allocation complexity
 
-inline __host__ __device__ int32_t compute_order(int32_t num_coeffs) {
+// Host-only: uses std::sqrt (host stdlib). All call sites construct host-side
+// device parameter structs or live in host launcher functions, so __device__
+// annotation is unnecessary and triggers stricter nvcc 12.9+ diagnostics.
+inline __host__ int32_t compute_order(int32_t num_coeffs) {
```
frtrudel: In our internal repo the signature has changed already. Preferable to avoid touching to avoid conflicts.

gsplat/cuda/csrc/RasterizeToPixelsFromWorldNHT3DGSFwd.cu
```diff
@@ -0,0 +1,710 @@
...
+        const at::optional<c10::intrusive_ptr<RowOffsetStructuredSpinningLidarModelParametersExt>> &, \
+        const at::optional<c10::intrusive_ptr<extdist::BivariateWindshieldModelParameters>> &, \
+        const at::Tensor, const at::Tensor,                                    \
+        bool, const at::Tensor, float,                                         \
+        const at::optional<at::Tensor>, bool,                                  \
+        at::Tensor, at::Tensor,                                                \
+        const at::optional<at::Tensor>, const at::optional<at::Tensor>,        \
+        at::Tensor);                                                           \
+
+__INS__(4, at::Half) __INS__(8, at::Half) __INS__(12, at::Half)
```
frtrudel, line 741: We don't support below CDIM=4? e.g. 3

gsplat/cuda/include/ExternalDistortion.cuh
Arcanous98: I can remove it but maybe it would be better if you could upstream the current signature? Otherwise I had issues compiling on CUDA 12.9

gsplat/cuda/csrc/Config.h
```diff
@@ -56,6 +57,10 @@
 #   define GSPLAT_BUILD_RELOC GSPLAT_DEFAULT_ENABLE_BUILD
 #endif
 
+#ifndef GSPLAT_BUILD_NHT
+#   define GSPLAT_BUILD_NHT GSPLAT_DEFAULT_ENABLE_BUILD
+#endif
+
 #ifndef GSPLAT_NUM_CHANNELS
-#   define GSPLAT_NUM_CHANNELS 1,2,3,4,5,8,9,16,17,24,32,33,64,65,128,129,256,257,512,513
+#   define GSPLAT_NUM_CHANNELS 1,2,3,4,5,8,9,12,16,17,24,32,33,48,64,65,80,96,128,129,256,257,512,513
```
frtrudel: A general thought here (perhaps outside the scope of this PR), until we have proper JIT'ing in place, it would be valuable to scope the values of this list per build target. i.e. 3DGUT is paying in compile time for new additions for NHT. Similarly dims only seen in 3DGS (e.g. 3dgs-specific unit tests) are shared to other targets here and vice-versa. Intermediate solution could be to append to that list based on enabled build targets.

gsplat/rendering.py
```diff
@@ -711,6 +722,7 @@ def check_features(features: Tensor, sh_degree: Optional[int], name: str) -> boo
         # Use provided UT parameters or create default
         if ut_params is None:
             ut_params = UnscentedTransformParameters()
+        projection_global_z_order = False if camera_model == "lidar" else global_z_order
```
frtrudel: That seems like a constraint that we'd want to at least document. For us it might not be a big deal since lidar is only supported with 3dgut and in our client code(s) global_z_order is always false, but other stakeholders might be passing global_z_order to true and that would get completely silenced.

examples/simple_trainer_nht.py
```diff
@@ -0,0 +1,1988 @@
...
+        for path in pdf_paths:
+            print(f"  - {path.name}")
+
+    @torch.no_grad()
+    def run_compression(self, step: int):
+        """Entry for running compression."""
+        print("Running compression...")
+        world_rank = self.world_rank
+
+        compress_dir = f"{cfg.result_dir}/compression/rank{world_rank}"
```
frtrudel: Static analysis showed that bare `cfg` is undefined here. Should be `self.cfg.result_dir`.

gsplat/cuda/csrc/RasterizeToPixelsFromWorldNHT3DGSFwd.cu
```diff
@@ -0,0 +1,710 @@
...
+    uint32_t cur_idx = 0;
+    uint32_t tr = block.thread_rank();
+
+    constexpr uint32_t OUT_DIM = constexpr_max(1U, CDIM * ENCF / 4);
+    float pix_out[OUT_DIM] = {0.f};
+    // Aux accumulators for fused depth / normals (zero-initialised, only
+    // updated when the corresponding output is requested).
+    float depth_out = 0.f;
+    vec3  normal_out = {0.f, 0.f, 0.f};
+#define ACC_ADD(idx, val, vis) do { acc_add_float(pix_out[idx], val, vis); } while(0)
```
frtrudel: Nit: Couldn't we have something else than an inline macro here that needs an undef later? Lambdas, separate inline device function, or just directly use acc_add_float?

gsplat/cuda/csrc/RasterizeToPixelsFromWorldNHT3DGSFwd.cu
```diff
@@ -0,0 +1,710 @@
...
+        } else {
+            inside = false;
+        }
+    } else {
+        i = block.group_index().y * tile_size + block.thread_index().y;
+        j = block.group_index().z * tile_size + block.thread_index().x;
+        inside = (i < image_height && j < image_width);
+    }
+
+    constexpr uint32_t OUT_CDIM = CDIM / VERTEX_PER_PRIM;
```
frtrudel, line 178: What about CDIMs that are not divisible by 4?

gsplat/cuda/include/HalfVectorLoads.cuh
```diff
@@ -0,0 +1,175 @@
...
+
+// Optional: set to 1 to enforce LD128/LD64 via inline PTX (single 128/64-bit load)
+#ifndef GSPLAT_ENFORCE_LD128_LD64_PTX
+#define GSPLAT_ENFORCE_LD128_LD64_PTX 0
+#endif
+
+namespace gsplat {
+
+// Load 8 halfs at ptr into out[0..7] as float using a single 128-bit load (LD128).
+// ptr must be 16-byte aligned for correct behavior.
```
frtrudel, line 43: `// ptr must be 16-byte aligned for correct behavior.` is there anything we could do to enforce that? Unless I missed them I don't see guards or alignas directives in code calling those.

gsplat/rendering.py
Arcanous98: global_z_order is required for correct lidar rendering unless we change the camera model. Right now gsplat uses a spherical convention for the coordinates so negative z are downward facing rays, which without this flag would get gaussians culled below the "horizon". Maybe we can do an assert instead of a silent set, but I would not allow global_z_order True when rendering lidars as I don't think it works.

frtrudel: I agree and I remember Rodolfo mentioning this. Making it more obvious via new assert would be good, because the only one we have AFAIK that's related to global_z_order is
```
    assert global_z_order or with_ut, "global_z_order can be false only if with_ut=True"
```
which isn't exactly gating the `camera_model=="lidar" and global_z_order` footgun.

gsplat/cuda/include/ExternalDistortion.cuh
frtrudel: Actually it seems the related PRs already had been ccreated, but never got merged in main due to missing dependencies. See https://github.com/nerfstudio-project/gsplat/pull/942

examples/simple_trainer_nht.py
Arcanous98: fixed :)

gsplat/cuda/csrc/RasterizeToPixelsFromWorldNHT3DGSFwd.cu
Arcanous98: I swapped to using acc_add_float

gsplat/cuda/csrc/RasterizeToPixelsFromWorldNHT3DGSFwd.cu
Arcanous98: Added an assert to make sure we don't try to use NHT without a CDIM divisible by 4. It would require a different virtual geometry for the interpolation. If we add 2DGS+NHT support in the future we can revisit it.

gsplat/cuda/include/HalfVectorLoads.cuh
Arcanous98: I added some guards

gsplat/rendering.py
Arcanous98: changed to an assert for NHT + lidar

gsplat/cuda/csrc/RasterizeToPixelsFromWorldNHT3DGSFwd.cu
Arcanous98: Not unless we change the virtual scaffold. It is a tetrahedron now i.e. requires 1 feature per vertex each

Arcanous98:
I rebased again and addressed the MR comments. Also, added batched inference mode for the MLP when rendering very high res (over 4k) images to avoid OOM.

Arcanous98:
I've added fused kernels for both the forward and backward passes. They compute in a single kernel both the rasterization pass (\w the tetrahedral interpolation and harmonic encoding) and the MLP deferred shading. Below are the results on 1M - 30k MipNeRF360 (averages over 9 scenes) in an RTX 4090:

| Split | N | Fwd Fused | Fwd TCNN | Fwd Spd | F+B Fused | F+B TCNN | F+B Spd |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| M360-Indoor | 4 | 12.44 ms | 20.44 ms | 1.64x | 57.30 ms | 64.85 ms | 1.13x |
| M360-Outdoor | 5 | 5.09 ms | 6.20 ms | 1.22x | 31.67 ms | 36.20 ms | 1.14x |
| Overall | 9 | 8.36 ms | 12.53 ms | 1.50x | 43.06 ms | 48.93 ms | 1.14x |

Trained quality is almost exactly the same, within ~1% of the old kernels. It's also possible to train with the non-fused (e.g. you want depth supervision) but eval with the fused one, as it produces the same result.

For now my plan is to keep both the separate and the fused kernels, as the new one is not feature complete (no depth gradients, no AOV mode) and I think for research it is easier to test things on the TCNN backend. But this opens the possibility to remove the TCNN dependency in the future.

gsplat/cuda/include/HalfVectorLoads.cuh
```diff
@@ -0,0 +1,195 @@
+/**
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
```
vince-brisebois, line 3: We'll want to have the full license headers, can you audit all the new files and make sure they have the full headers? See examples/simple_trainer.py for example.

examples/simple_trainer_nht.py
vince-brisebois, line 1: Not sure if this is beyond what you could give in terms of bandwidth, but I would love to see us create a GaussianNHTScene (see gsplat/scene) and then dispatch rendering via:
https://github.com/nerfstudio-project/gsplat/blob/d28ee0c42264a540b8b17388cf793ad31bc20b53/gsplat/experimental/render/functional/render_scene.py#L33

Arcanous98:
I fixed the licenses in all new files, rebased on the newest commits and added `GaussianNHTScene `. Also fixed some windows specific compilation errors. Let me know if everything is ready for the merge :)

Arcanous98:
Rebased again and improved fused kernel performance.

PR #955, opened 2026-05-24
Port gsplat to AMD HIP/ROCm

jungx098:
Adds HIP/ROCm support so all 10 kernel modules (3DGS, 2DGS, 3DGUT, SH, Adam, Relocation, CameraWrappers, ExtDistWrappers, IntersectTileLidar, Null) build and run on AMD GPUs via PyTorch's HIP backend. Every change is `#ifdef USE_ROCM`-gated, or is unconditional and verified to preserve CUDA semantics. Tested on AMD RX 9070 XT (gfx1201, RDNA4) under ROCm 7.13.0 / PyTorch 2.11+rocm7.13.

### Changes

- **`build.py`**: hipify integration, HIP-conditional include paths, GLM source-patch (`__CUDACC__` → `!__HIP__`), 8× `cudaFuncSetAttribute` `(const void*)` casts, `csrc/*.{h,cuh}` mirror.
- **Kernels** (`#ifdef USE_ROCM`): 8× `cg::labeled_partition` + warpSum → per-thread `gpuAtomicAdd`; 3× `cg::reduce` → `_warp_butterfly_max` helper; `cub` → `hipcub` alias; new `CudaStdOptional.h` shim — ROCm's `<cuda/std/optional>` is broken under hipcc (NVIDIA-only fp8 types in its transitive includes) and host `<optional>` throws from `__host__ __device__` code.
- **Headers**: `Cameras.cuh` (`min`/`max` → `std::fmin`/`fmax`, `.at()` → `[]`), `Utils.cuh` (butterfly reductions), `TensorView.h` (device-arch guard widened to `__HIP_DEVICE_COMPILE__`).
- **Reference impl** (`_torch_impl_ut.py`): closed-form 2×2 inverse for `cov_2d` — dodges a torch+ROCm LU crash and is faster on every backend.
- **Test infra** (`conftest.py`): `dist_init` NCCL warm-up in try/except so HIP boxes without RCCL collectives don't break unrelated tests.

### Tests

Per-module: SH 15/0, strategy 2/0, cameras+extdist 174/0, 2DGS 12/0, 3DGUT+isect_lidar 108/0. Full suite **597 pass / 9 fail / 30 skip** (RX 9070 XT, ROCm 7.13). The 9 fails are all out-of-scope: 6× near-singular-input tolerance on rolling-shutter unscented-transform (`test_fully_fused_projection_ut`) tests, 1× per-thread-atomic lidar variant 1.5% over atol, 1× `-use_fast_math` `atan2` precision (1-in-1M fisheye gradient), 1× `test_compression` blocked on an unrelated cupy-rocm bug (fix filed upstream as [PR #9961](https://github.com/cupy/cupy/pull/9961)).

End-to-end on Mip-NeRF 360 Bonsai (30k steps): **2DGS PSNR 32.51 in 15:55; 3DGS PSNR 32.40 in 9:32** — both match the upstream CUDA reference.

### Caveats

- Per-thread atomics replace warp-reduced atomics in the HIP path (perf cost not quantified).
- `wave32` hardcoded in 3 `cg::tiled_partition<32>` sites — fine on RDNA3/4; wave64/CDNA would need parameterisation.
- 9 rasterizer tests need `nerfacc` to be importable on HIP — currently blocked because nerfacc's CUDA extension doesn't build on ROCm. Out of scope for this PR; the right fix is an upstream PR to `nerfstudio-project/nerfacc`.
- The 3DGS example trainer (`simple_trainer.py`) also uses [`rahul-goel/fused-ssim`](https://github.com/rahul-goel/fused-ssim) for its loss; building it on dual-CUDA+ROCm hosts needs a one-line `-isystem` workaround (separate concern; not required by this PR or by the kernels themselves).

### Reviewer note

Happy to split into smaller PRs if preferred. No CUDA box to re-verify the CUDA test suite — all HIP-specific changes are `#ifdef USE_ROCM`-gated; the unconditional patches (TensorView guard, `.at()` → `[]`, closed-form 2×2 inv) preserve CUDA semantics.

### DIFF
--- a/gsplat/cuda/_torch_impl_ut.py
+++ b/gsplat/cuda/_torch_impl_ut.py
@@ -513,15 +513,23 @@ def _fully_fused_projection_with_ut(
         valid_gaussian & (cov_diag_check[..., 0] > 0.0) & (cov_diag_check[..., 1] > 0.0)
     )
 
-    # Compute conics (inverse of 2D covariance)
-    # This is more robust than manual formula, especially for non-symmetric matrices
-    # (numerical errors in weighted sum can break exact symmetry)
-    # Add a small epsilon to the diagonal to prevent torch.linalg.inv from
-    # producing NaN on singular matrices (invalid Gaussians are masked out
-    # by valid_gaussian anyway, but autograd still propagates through inv).
-    cov_2d_inv = torch.linalg.inv(
-        cov_2d + 1e-6 * torch.eye(2, dtype=cov_2d.dtype, device=cov_2d.device)
+    # 2x2 inverse via closed form: M^-1 = (1/det) * [[d,-b],[-c,a]].
+    # Epsilon on the diagonal avoids div-by-zero on near-singular invalid
+    # Gaussians (autograd still propagates through the inverse).
+    # Avoid torch.linalg.inv: torch+ROCm's batched 2x2 LU kernel raises
+    # hipErrorInvalidConfiguration on non-finite inputs (gfx1201, torch
+    # 2.11+rocm7.13). Closed form sidesteps it and is faster for 2x2.
+    cov_2d_blur = cov_2d + 1e-6 * torch.eye(
+        2, dtype=cov_2d.dtype, device=cov_2d.device
     )  # [B, C, N, 2, 2]
+    a = cov_2d_blur[..., 0, 0]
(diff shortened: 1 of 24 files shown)


HPC-Ken:
After discussing with @jungx098, PR #957 is our preferred contribution.

PR #958, opened 2026-05-29
rays reshape fix

AlisherBlack:
The eval3d pixel rasterizer (rasterize_to_pixels_eval3d_extra) expects per-ray input flattened as (..., C, H*W, 6): it derives P = rays.shape[-2] and its assert_shape wants rank (num_batch_dims + 3). The public API above takes (and asserts) the image-shaped (..., C, H, W, 6), so flatten the H, W dims here. Row-major flattening matches the kernel's pixel index (row * W + col) and the image-shaped output, so no reshape is needed on the returned colors/alphas.

https://github.com/nerfstudio-project/gsplat/issues/967

### DIFF
--- a/gsplat/rendering.py
+++ b/gsplat/rendering.py
@@ -1108,6 +1108,15 @@ def rasterization(
         }
     )
 
+    # The eval3d pixel rasterizer (rasterize_to_pixels_eval3d_extra) expects per-ray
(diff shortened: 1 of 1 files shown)


AlisherBlack:
@rodlima78, @vince-brisebois, could you plz review?

PR #1028, opened 2026-07-15
Fix cudaEventCreate build error in Utils.cpp merge_streams

adelpit:
## Problem

`merge_streams()` in `gsplat/cuda/csrc/Utils.cpp` (added in #1019  / commit 4561ac47) calls `cudaEventCreate` with two arguments:

```cpp
C10_CUDA_CHECK(cudaEventCreate(&events[device_id], cudaEventDisableTiming));
```

The C runtime function `cudaEventCreate` takes a single argument. The two-argument form is a C++ convenience overload declared in `cuda_runtime.h`, which is not in scope here: `Utils.cpp` is compiled by the host C++ compiler and only pulls in `cuda_runtime_api.h` (directly and via the c10 CUDA headers). In environments where `cuda_runtime.h` isn't otherwise dragged in, the build fails:

```
gsplat/cuda/csrc/Utils.cpp:39:39: error: too many arguments to function 'cudaError_t cudaEventCreate(CUevent_st**)'
```

(Observed building with pip from source on CUDA 12.9 / torch 2.x / gcc, Ubuntu 22.04. The same pattern works in `.cu` files because nvcc auto-includes `cuda_runtime.h`.)

## Fix

Use `cudaEventCreateWithFlags`, the C API function intended for passing `cudaEventDisableTiming`. No behavior change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/gsplat/cuda/csrc/Utils.cpp
+++ b/gsplat/cuda/csrc/Utils.cpp
@@ -36,12 +36,12 @@ void merge_streams()
     {
         C10_CUDA_CHECK(cudaSetDevice(device_id));
         auto stream = c10::cuda::getCurrentCUDAStream(device_id);
-        C10_CUDA_CHECK(cudaEventCreate(&events[device_id], cudaEventDisableTiming));
+        C10_CUDA_CHECK(cudaEventCreateWithFlags(&events[device_id], cudaEventDisableTiming));
         C10_CUDA_CHECK(cudaEventRecord(events[device_id], stream));
     }
 
     C10_CUDA_CHECK(cudaSetDevice(merge_device_id));
-    C10_CUDA_CHECK(cudaEventCreate(&merge_event, cudaEventDisableTiming));
+    C10_CUDA_CHECK(cudaEventCreateWithFlags(&merge_event, cudaEventDisableTiming));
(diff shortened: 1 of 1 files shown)


Arison591:
Tested this fix on a real Blackwell consumer GPU.

Environment:

- NVIDIA GeForce RTX 5090 (compute capability 12.0), driver 580.76.05
- CUDA toolkit 12.8.93
- PyTorch 2.7.0+cu128, Python 3.12

On current `main`, I reproduced the reported host compilation error in `Utils.cpp`:

```text
error: too many arguments to function ‘cudaError_t cudaEventCreate(CUevent_st**)’
```

After applying this PR's two `cudaEventCreateWithFlags` substitutions, I ran:

```bash
PATH=/usr/local/cuda/bin:$PATH \
MAX_JOBS=4 \
TORCH_CUDA_ARCH_LIST=12.0 \
python -m pytest tests/test_strategy.py -q -s
```

The JIT extension compiled and linked with `-gencode=arch=compute_120,code=sm_120`, loaded successfully, and all 5 tests passed. This included CUDA rasterization forward/backward and both strategy paths. The patch fixes the SM120 source build without changing runtime behavior.

PR #1032, opened 2026-07-22
Add weighted-median LiDAR return rendering

whao22:
## Summary
- add an opt-in `lidar_return_mode="median"` for LiDAR Eval3D hit-distance rendering
- select the hit distance whose cumulative opacity crosses 0.5, avoiding expected-distance interpolation across depth discontinuities
- constrain median returns to inference-only MixedBatch rendering and preserve existing expected-return behavior by default
- add Python validation/regression coverage and update native Eval3D call sites

## Testing
- `conda run -n gsplat2 env PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -o addopts='' tests/test_basic.py::test_rasterization_validates_lidar_return_mode tests/test_basic.py::test_rasterization_lidar_median_selects_threshold_crossing_hit -q` (3 passed)
- `conda run -n gsplat2 env PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -o addopts='' tests/test_cpp.py -q` (124 passed)

### DIFF
--- a/gsplat/__init__.py
+++ b/gsplat/__init__.py
@@ -62,6 +62,7 @@ from .cuda._wrapper import (
 from .exporter import export_splats
 from .optimizers import SelectiveAdam
 from .rendering import (
+    LidarReturnMode,
     RasterizeMode,
     RendererConfig,
     RendererConfig_MixedBatch,
@@ -111,6 +112,7 @@ __all__ = [
     "CameraModel",
     "ExternalDistortionModelMeta",
     "ExternalDistortionModelParameters",
+    "LidarReturnMode",
     "RasterizeMode",
     "RendererConfig",
     "RendererConfig_MixedBatch",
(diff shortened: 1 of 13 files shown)


whao22:
### Motivation

This PR is intended to address the radial/scattered point artifacts that can appear around depth discontinuities when rendering LiDAR with the expected hit-distance mode (`Ed`). The expected return averages the hit distances of all contributing Gaussians along a ray. Near object boundaries such as wall corners, foreground and background Gaussians can overlap in screen space, producing non-physical intermediate distances that appear as rays extending outward from the edges.

The following screenshots show representative artifacts from the expected-distance output (the second image highlights one affected region in red):

Radial edge artifacts in expected LiDAR rendering

Highlighted radial edge artifacts in expected LiDAR rendering

The new weighted-median return selects the hit distance at which cumulative opacity first crosses 0.5 instead of averaging distances across separated surfaces. This avoids generating intermediate returns between foreground and background geometry and better matches LiDAR first-return behavior. The existing expected mode remains the default for backward compatibility.

PR #1057, opened 2026-09-10
[perf] 3DGS training (simple_trainer.py) on 8x B200: 14.1 ms to 9.3 ms per step, mostly from decoding one image per step

TarzanZhao:
## Summary

Training 3DGS with `examples/simple_trainer.py` on 8 B200 GPUs, a good part of the 14.1 ms step was not rendering. Every step waited for a DataLoader worker to decode one PNG and then copied it to the GPU as 13 MB of float32, and the densification statistics and the progress text each read a value back from the device, which makes the CPU wait for the step it just queued. This PR adds an opt-in image cache and removes the two reads. The step goes from 14.1 ms to 9.3 ms (**1.51x**) with `--cache_images`, and to 12.9 ms (1.09x) without it. The whole run stays at about 74 s, because its startup dominates and this PR does not touch it. I have not measured other GPUs or scenes.

1. **`--cache_images`**, off by default, COLMAP data only. Upstream decodes and undistorts an image on every access and hands it over as float32: 13.07 MB for a 1297x840 image, 91.5 GB over the 875 steps of the 8 ranks. With the flag, the dataset decodes each training image once before training and keeps it as uint8 in host memory. A step then moves 3.27 MB, 22.9 GB over the run, and the divide by 255 happens on the GPU, so the pixel values are the same. The DataLoader already pinned every batch, but the copy to the GPU was blocking; it is now issued with `non_blocking=True`. The crop, the depth points and the rest of the sample are still built per access, so `--patch_size` and `--depth_loss` work with it (covered by the CPU tests below; the benchmark scene uses neither, so they were not measured on the GPU). The cost is the decoded training set in host memory on every rank, about 3 MB per image at this resolution. It is rejected for NCore data, which loads through its own dataset.
2. **Dense running statistics without a host sync**. `DefaultStrategy._update_state` selected the visible Gaussians with `torch.where` and boolean indexing, so the step waited for the device to know how many there were before it could `index_add_`. Masked reductions over the camera axis keep the shapes fixed and remove the wait. One behaviour change comes with it: with more than one camera per step, the stored radius is now the true maximum over the cameras, where the old indexed write kept one arbitrary camera's radius. That path is live only under `refine_scale2d_stop_iter > 0`, which is 0 by default. The packed path is unchanged.
3. **Progress text every 10 steps**. Building the tqdm description calls `loss.item()`, which waits for the backward to finish before the CPU can queue the next step. It is now refreshed every tenth step and on the last one, so the displayed loss can be up to nine steps old.
4. **Viewer imports only when the viewer is used**. `viser` and `nerfview` were imported at module import even under `--disable_viewer`, which every distributed run uses.

### Speedup Result

Measured on 8 B200 GPUs, two runs per version, interleaved.

| | unmodified `28e794ca` | this PR, `--cache_images` | change |
|---|---|---|---|
| training step, mean of steps 100 to 874 | 14.02 / 14.26 ms | 9.37 / 9.32 ms | **1.51x** |
| training step, median | 10.42 / 10.35 ms | 8.26 / 8.10 ms | 1.27x |
| training step, cache off (items 2 to 4 only) | | 12.93 ms mean, 9.82 ms median | 1.09x |
| training loop, 875 steps | 13.25 / 13.31 s | 10.76 / 8.80 s | |
| whole run, launch to exit | 74.1 / 74.7 s | 71.3 / 77.1 s | unchanged |
| peak GPU memory per rank | 0.97 / 0.97 GB | 0.96 / 0.95 GB | |
| host memory per rank | 2.51 / 2.52 GB | 2.99 / 2.90 GB | +0.43 GB, the decoded images |
| one-time cost before the loop | | about 1.5 s per rank | decoding 161 images |

The two loop times on the right differ because the first cache run spent 2.86 s in step 0 against 0.93 s in the second, rank skew at the start of the loop; the steady step is the same in both.

`pytest tests/test_strategy.py tests/test_dataset_colmap.py` with `CUDA_VISIBLE_DEVICES=""`: 9 passed, 2 skipped (the two CUDA-only tests that were already there). Three of the new strategy cases fail on unmodified `28e794ca`, exactly on the multi-camera radius maximum described above.

## Correctness Verification

I trained the unmodified code and this branch on the same scene with the same seed, and compared 9016 recorded values, 1127 on each of the 8 ranks: the id of the image every one of the 875 steps trained on, and at steps 0 to 9, every 50th step and the last step, the loss, its L1 and SSIM parts, the mean and maximum of the rendered image, the number of visible Gaussians, the Gaussian count after refinement, and the render and alpha shapes. **All 9016 are within tolerance on all 8 ranks.** Step 0 is bit-identical (loss 0.26062461733818054 on both), and every step trained on the same image on both sides, so this compares the two runs step by step rather than two trajectories. The closest call in the table is the L1 part of the loss at step 350 on rank 7, at 0.64 of its tolerance. At the end of the run, rank 0 has loss 0.060503 unmodified and 0.060025 on this branch.

The tolerances in the table were set before the runs, each one three times the spread the same checkpoint shows over repeats of the unmodified code. That spread is not rounding: the rasterizer's backward accumulates gradients atomically, so two runs of the same code already drift from step 1, and once refinement starts they keep different Gaussians. The two rows on the Gaussians separate the two kinds of difference: while refinement has not started the counts are identical, and what grows apart afterwards is which Gaussians each run kept, not the arithmetic of a step. The recording code is on the `perf-gsplat-train-8gpu-verify` branch of my fork `TarzanZhao/gsplat-perf`, not in this PR.

| recorded | unmodified vs this branch | tolerance (step 0 / steps 1 to 9 / later) |
|---|---|---|
| image id of the step, all 875 steps x 8 ranks | identical | exact |
| loss, L1, SSIM, and the mean of \|rgb\| | within tolerance on all ranks | exact / 0.03 % / 6 % |
| visible Gaussians at the recorded steps | identical at step 0; at most 4 differ of the 126000 to 139000 visible at steps 1 to 9; up to 3.5 % later | exact / 0.01 % / 20 % |
| Gaussian count after refinement | identical through step 50, before refinement starts; at the last step between 5.2 % fewer and 6.1 % more, by rank, the widest 295300 against 313204 | exact / 0.01 % / 20 % |
| render and alpha shapes | identical | exact |

<details>
<summary>Details: hardware, model, full command</summary>

**Hardware.** One node, 8 x NVIDIA B200 (sm_100, 183 GB each), 2 x Xeon 8562Y+ (128 threads), driver 580.126.20, CUDA 12.8. Python 3.11, torch 2.9.1+cu128, NCCL 2.27.5, gsplat 1.6.0 built from source (`pip install -e . --no-build-isolation`, `TORCH_CUDA_ARCH_LIST=10.0`), fused-ssim 1.0.0, imageio 2.37.4, Pillow 12.3.0, pycolmap 4.2.0, torchmetrics 1.8.2. The scene sits on node-local disk.

**Model.** 3D Gaussian Splatting of the Mip-NeRF 360 `garden` scene, 161 training images at 1297 x 840. Gaussians start from 138766 COLMAP points and grow to about 2.6 M (330 k per rank), sharded across the 8 ranks; SH degree 3 from loop step 375, six `torch.optim.Adam(fused=True)` optimizers, fp32 rasterization, `DefaultStrategy` with `refine_start_iter` 62, `refine_every` 12 and `reset_every` 375 after `adjust_steps(0.125)`, one camera per rank, L1 + SSIM loss. Seed 42 + rank, set by the script.

**Measurement.** Against the unmodified `28e794ca`, the script as shipped, two runs per version interleaved (base, this PR, base, this PR), each alone on the node with all 8 GPUs idle beforehand. Both versions export `NCCL_CUMEM_ENABLE=0` and `PYTORCH_NVML_BASED_CUDA_CHECK=1`, which are not part of this PR: the first takes the first step of the unmodified code from 9 to 12 s down to 1 to 5 s on this node, the second is within noise. Step time comes from a `perf_counter` tick at the top of every step on rank 0, recorded by a `sitecustomize.py` outside the repo; the script's own `ellipse_time` agrees with the summed ticks. Steps 0 to 99 are excluded from the mean because step 0 waits on the slowest rank and refinement starts at step 72. Run-to-run spread over these runs: 0.24 ms on the step, up to 6 s on the wall time.

**Data** (once): `python examples/datasets/download_dataset.py --dataset mipnerf360 --save_dir <data>`, then copy `<data>/360_v2/garden` to local disk.

**The measured job**, both versions, with `--cache_images` added on this branch only:
```
export NCCL_CUMEM_ENABLE=0
export PYTORCH_NVML_BASED_CUDA_CHECK=1
CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 python examples/simple_trainer.py default \
  --disable_viewer --data_dir <local>/garden/ --data_factor 4 --result_dir <out> \
  --max_steps 7000 --steps_scaler 0.125 \
  --eval_steps -1 --save_steps -1 --ply_steps -1 --disable_video --tb_every 0 --port 31200 \
  --cache_images     # this branch only
```
`--steps_scaler 0.125` is the repo's multi-GPU setting from the `simple_trainer.py` docstring: 875 loop steps of 8 images each, with the refine and reset schedules scaled to match. The training loop time is `ellipse_time` in `<out>/stats/train_step0874_rank0.json`; it starts after the cache is built, so the whole-run row above is the wall time of the command.

**The cache and the loader.** `preload_images` runs before the loader starts its workers, so the four workers share the cache of the process they are forked from; the loader, its shuffling and its pinning are untouched, which is why the image order is identical. Where the default start method is spawn rather than fork, each worker would get its own copy of the cache; I measured Linux only. The dataset returns uint8 only after `preload_images`, which nothing but `--cache_images` calls, so other users of `datasets/colmap.py` see float32 as before. Serving the cache from a zero-worker loader instead was 4 to 7 times slower than upstream (69 ms and 41 ms per step): collating and pinning a 3 MB image in the trainer process uses all 128 threads of the node, in 8 ranks at once. DataLoader workers set their thread count to 1, so leaving the loader alone avoids this.

**Settings tried and rejected** on the unmodified code: `OMP_NUM_THREADS=16`, `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`, `NCCL_NVLS_ENABLE=0` and `NCCL_IB_DISABLE=1` were within noise, and `--packed` was slower (median step +3.3 ms). TF32, `--sparse_grad`, strategy settings and `--antialiased` were not tried because they change the numerics or the densification schedule.

</details>


TarzanZhao:
This PR is ready for review. It speeds up 3DGS training in simple_trainer.py on 8 B200 GPUs: step time 13.9 ms to 7.9 ms (1.8x), whole run 2.5x.

cc @liruilong940607 (last change to simple_trainer.py and merged the last PR touching it, original author of distributed.py), @vince-brisebois (last change to strategy/default.py, recent changes to distributed.py), @matthewdcong (last change to distributed.py). Any feedback is welcome.

TarzanZhao:
I've cleaned this PR up and cut it down a lot.

@vince-brisebois any chance you could take a look, or is there someone better you'd suggest?

TarzanZhao:
I have picked the essential part of this PR into a smaller one: #1062, the image cache alone, measured on its own. Moving this one back to draft.

TarzanZhao:
Closing this since #1062 is merged.

